PR #6, opened 2021-03-08
finish AlignmentInspector

souporserious:
Finishes alignment inspector by consolidating `align` into one action as well as adding a `distribute` action. This now utilizes `AffineTransform` to properly space nested layers. We'll need to account for some weird cases as ExcaliDraw does [here](https://github.com/excalidraw/excalidraw/blob/master/src/disitrubte.ts) and make sure to disable buttons when an alignment/distribution cannot be made (e.g. you can't distribute less than 3 layers).

align

### DIFF
--- a/packages/app/src/components/inspector/AlignmentInspector.tsx
+++ b/packages/app/src/components/inspector/AlignmentInspector.tsx
@@ -27,16 +27,20 @@ function AlignmentInspector(props: AlignmentInspectorProps) {
   const [, dispatch] = useApplicationState();
   return (
     <AlignmentInspectorContainer>
-      <SpaceEvenlyHorizontallyIcon />
-      <SpaceEvenlyVerticallyIcon />
+      <SpaceEvenlyHorizontallyIcon
+        onClick={() => dispatch('distribute', 'horizontal')}
+      />
+      <SpaceEvenlyVerticallyIcon
+        onClick={() => dispatch('distribute', 'vertical')}
+      />
       <AlignLeftIcon onClick={() => dispatch('align', 'left')} />
       <AlignCenterHorizontallyIcon
-        onClick={() => dispatch('align', 'centerHorizontally')}
+        onClick={() => dispatch('align', 'centerHorizontal')}
       />
       <AlignRightIcon onClick={() => dispatch('align', 'right')} />
       <AlignTopIcon onClick={() => dispatch('align', 'top')} />
       <AlignCenterVerticallyIcon
-        onClick={() => dispatch('align', 'centerVertically')}
+        onClick={() => dispatch('align', 'centerVertical')}
       />
       <AlignBottomIcon onClick={() => dispatch('align', 'bottom')} />
     </AlignmentInspectorContainer>
--- a/packages/app/src/components/inspector/AlignmentInspector.tsx
+++ b/packages/app/src/components/inspector/AlignmentInspector.tsx
@@ -30,10 +30,14 @@ function AlignmentInspector(props: AlignmentInspectorProps) {
       <SpaceEvenlyHorizontallyIcon />
       <SpaceEvenlyVerticallyIcon />
       <AlignLeftIcon onClick={() => dispatch('alignLeft')} />
-      <AlignCenterHorizontallyIcon />
+      <AlignCenterHorizontallyIcon
+        onClick={() => dispatch('alignCenterHorizontally')}
+      />
       <AlignRightIcon onClick={() => dispatch('alignRight')} />
       <AlignTopIcon onClick={() => dispatch('alignTop')} />
-      <AlignCenterVerticallyIcon />
+      <AlignCenterVerticallyIcon
+        onClick={() => dispatch('alignCenterVertically')}
+      />
       <AlignBottomIcon onClick={() => dispatch('alignBottom')} />
     </AlignmentInspectorContainer>
   );
--- a/packages/app/src/components/inspector/AlignmentInspector.tsx
+++ b/packages/app/src/components/inspector/AlignmentInspector.tsx
@@ -29,16 +29,16 @@ function AlignmentInspector(props: AlignmentInspectorProps) {
     <AlignmentInspectorContainer>
       <SpaceEvenlyHorizontallyIcon />
       <SpaceEvenlyVerticallyIcon />
-      <AlignLeftIcon onClick={() => dispatch('alignLeft')} />
+      <AlignLeftIcon onClick={() => dispatch('align', 'left')} />
       <AlignCenterHorizontallyIcon
-        onClick={() => dispatch('alignCenterHorizontally')}
+        onClick={() => dispatch('align', 'centerHorizontally')}
       />
-      <AlignRightIcon onClick={() => dispatch('alignRight')} />
-      <AlignTopIcon onClick={() => dispatch('alignTop')} />
+      <AlignRightIcon onClick={() => dispatch('align', 'right')} />
+      <AlignTopIcon onClick={() => dispatch('align', 'top')} />
       <AlignCenterVerticallyIcon
-        onClick={() => dispatch('alignCenterVertically')}
+        onClick={() => dispatch('align', 'centerVertically')}
       />
-      <AlignBottomIcon onClick={() => dispatch('alignBottom')} />
+      <AlignBottomIcon onClick={() => dispatch('align', 'bottom')} />
     </AlignmentInspectorContainer>
   );
 }
--- a/packages/noya-state/src/reducers/application.ts
+++ b/packages/noya-state/src/reducers/application.ts
@@ -19,6 +19,7 @@ import {
   getCurrentPageMetadata,
   getSelectedLayerIndexPaths,
   getSelectedLayerIndexPathsExcludingDescendants,
+  getSelectedRect,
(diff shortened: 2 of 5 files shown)


packages/app/src/components/inspector/AlignmentInspector.tsx
```diff
@@ -27,14 +27,22 @@ function AlignmentInspector(props: AlignmentInspectorProps) {
...
     <AlignmentInspectorContainer>
-      <SpaceEvenlyHorizontallyIcon />
-      <SpaceEvenlyVerticallyIcon />
-      <AlignLeftIcon onClick={() => dispatch('alignLeft')} />
-      <AlignCenterHorizontallyIcon />
-      <AlignRightIcon onClick={() => dispatch('alignRight')} />
-      <AlignTopIcon onClick={() => dispatch('alignTop')} />
-      <AlignCenterVerticallyIcon />
-      <AlignBottomIcon onClick={() => dispatch('alignBottom')} />
+      <SpaceEvenlyHorizontallyIcon
```
dabbott, line 30: Let's move this file to `/containers`, since it uses `dispatch`. I think originally I had planned to expose a bunch of callbacks as props, and that's why this component is in `/components`. But I like this approach better, so let's just move the file

packages/noya-state/src/reducers/application.ts
```diff
@@ -73,12 +77,17 @@ export type Action =
   | [type: 'highlightLayer', highlight: LayerHighlight | undefined]
   | [type: 'setExpandedInLayerList', layerId: string, expanded: boolean]
   | [type: 'selectPage', pageId: UUID]
-  | [type: 'alignLeft']
-  | [type: 'alignCenterHorizontally']
-  | [type: 'alignRight']
-  | [type: 'alignTop']
-  | [type: 'alignCenterVertically']
-  | [type: 'alignBottom']
+  | [type: 'distribute', placement: 'horizontal' | 'vertical']
```
dabbott, line 80: Can we rename the actions to `distributeLayers` and `alignLayers`?

I'd like most action names to follow the format `verbObject`, like `selectPage` or `highlightLayer`. Some of them probably don't right now... but now that they're growing a bit, I'll try to keep them mostly to that format

packages/noya-state/src/reducers/application.ts
```diff
@@ -249,59 +258,147 @@ export function reducer(
...
     }
-    case 'alignLeft': {
+    case 'distribute': {
       const pageIndex = getCurrentPageIndex(state);
       const layerIndexPaths = getSelectedLayerIndexPaths(state);
+      const axis = action[1];
 
       return produce(state, (state) => {
+        const page = state.sketch.pages[pageIndex];
+        const selectedRect = getSelectedRect(layerIndexPaths, page);
```
dabbott, line 268: Let's calculate the `selectedRect` outside of `produce`. 

`getSelectedRect` calls `getBoundingRect`, which traverses every layer

packages/noya-state/src/reducers/application.ts
```diff
@@ -249,59 +258,147 @@ export function reducer(
...
-          (layerA, layerB) => layerA.frame.y - layerB.frame.y,
-        );
-        layers.forEach((layer) => {
-          layer.frame.y = topmostLayer.frame.y;
+        const differenceWidth = selectedRect.width - combinedWidths;
+        const differenceHeight = selectedRect.height - combinedHeights;
+        const gapX = differenceWidth / (layers.length - 1);
+        const gapY = differenceHeight / (layers.length - 1);
+        const sortBy = axis === 'horizontal' ? 'minX' : 'minY'; // TODO: sort by mid point once it's available
+        const sortedLayerIndexPaths = layerIndexPaths
```
dabbott, line 283: The input to this `map(...).sort(...).map(...)` operation is type `IndexPath[]`, and the output is also `IndexPath[]`. I'm wondering if we can get away with just a `sort(...)` and no maps?

```ts
const getNormalizedBounds = (indexPath: IndexPath): Bounds => {
  const layer = Layers.access(page, layerIndexPath);
  const transform = getNormalizedTransform(page, layerIndexPath);
  return Primitives.createBounds(
    Primitives.transformRect(layer.frame, transform)
  )
}

const sortedLayerIndexPaths = layerIndexPaths
  .sort((a, b) => getNormalizedBounds(a)[sortBy] - getNormalizedBounds(b)[sortBy]);
```

We also probably want to use the `Primitives.transformRect` helper for applying a transform to a layer. Otherwise we won't transform the width/height correctly (which doesn't happen yet, but I think if a symbol is scaled?). 

We could also make a `Primitives.transformBounds` to avoid a small amount of extra work here, but not too important

We could always memoize `getBounds` if needed for perf, but also probably not needed

packages/noya-state/src/reducers/application.ts
```diff
@@ -249,59 +258,147 @@ export function reducer(
         state.selectedPage = action[1];
       });
     }
-    case 'alignLeft': {
+    case 'distribute': {
       const pageIndex = getCurrentPageIndex(state);
       const layerIndexPaths = getSelectedLayerIndexPaths(state);
+      const axis = action[1];
```
dabbott, line 264: Minor: if I'm going to rename an action parameter, I usually do 

```ts
const [, axis] = action;
```

since this pattern works with any number of parameters

packages/noya-state/src/reducers/application.ts
```diff
@@ -249,59 +258,147 @@ export function reducer(
...
 
       return produce(state, (state) => {
+        const page = state.sketch.pages[pageIndex];
+        const selectedRect = getSelectedRect(layerIndexPaths, page);
         const layers = accessPageLayers(state, pageIndex, layerIndexPaths);
-        const [leftmostLayer] = layers.sort(
-          (layerA, layerB) => layerA.frame.x - layerB.frame.x,
+        const combinedWidths = layers.reduce(
```
dabbott, line 270: There may be a scaling issue here (which I describe below on `sortedLayerIndexPaths`) but I don't think we need to handle it yet

packages/noya-state/src/reducers/application.ts
```diff
@@ -249,59 +258,147 @@ export function reducer(
...
 
       return produce(state, (state) => {
-        const layers = accessPageLayers(state, pageIndex, layerIndexPaths);
-        const [bottommostLayer] = layers.sort(
-          (layerA, layerB) => layerB.frame.y - layerA.frame.y,
-        );
-        layers.forEach((layer) => {
-          layer.frame.y = bottommostLayer.frame.y;
+        const page = state.sketch.pages[pageIndex];
+        const selectedRect = getSelectedRect(layerIndexPaths, page);
```
dabbott, line 340: Same as above, move this outside `produce`

packages/noya-state/src/reducers/application.ts
```diff
@@ -874,6 +971,23 @@ function accessPageLayers(
   });
 }
 
+function getNormalizedTransform(page: Sketch.Page, indexPath: IndexPath) {
```
dabbott, line 974: Maybe move these helpers to selectors? 

For `getSelectedRect`, we could probably make `state: ApplicationState` the only parameter. I don't think we have a use case for doing this on a non-selected page

dabbott: I wonder if instead of "normalized" we should call this `getLayerTransform`? I'm hesitant to call the page's coordinate system the "normalized" one, only cause I'm not sure if that's an appropriate usage of "normalized" in matrix terms. 

Also, this could be called with any parent layer and index path, not just a page, and it will still make sense: it calculates the transform of the index path parameter relative to the layer parameter.

dabbott (changes requested):
I think this came out nicely - it's pretty concise for the amount of functionality. I have some misc feedback about naming and moving things around, but all minor stuff.

packages/noya-state/src/reducers/application.ts
souporserious: Oops yes, not enough info there 😅  fixed!

packages/noya-state/src/reducers/application.ts
souporserious: I was doing this originally because we write to layer below, I added a comment in the updated code, maybe pulling it out into a util could help document what it's doing better, something like `getWriteableLayer`?

packages/noya-state/src/reducers/application.ts
souporserious: So much nicer now! ✨

packages/noya-state/src/reducers/application.ts
souporserious: Ah, that makes sense. Cleaned that up so it's a little more general and better naming. Should `getNormalizedBounds` be renamed as well?

packages/noya-state/src/reducers/application.ts
dabbott: I think I see what you mean, how it's not clear whether a value is coming from the `state` within produce or from outside

I'm not sure what the best solution is here, but I agree we should make it more clear. I think we could also propagate the `Draft<T>` type through most functions by using generics, which would help, since then it would be enforced at the type level, to some extent (maybe?)

dabbott:
Looks great 🔥 , merging!

PR #7, opened 2021-03-13
Color swatches

### LINKED ISSUE #5: Shared color swatches
## Feature

Each document has a set of shared colors, i.e. `document.sharedSwatches`. In the "Components" tab within `Workspace`, we'll have:
- A grid view to display all shared colors
- A button in the toolbar to add a new color
- A way to edit existing colors by clicking to select and choosing a new color from the inspector

## Possible implementation

- Create a `GridView` in the `noya-designsystem` package. 
  - The API could be similar to `ListView`
  - We should support multiple selection just like `ListView`
  - We *don't* need to implement drag and drop yet
  - We *don't* need a callback for deleting items yet
- Create a `ColorSwatch` component in `app` package, maybe `components/swatches/ColorSwatch.tsx`
- Create a `ColorSwatchGrid` in `app` package, in `containers`
  - Get the document's colors from a helper function in `selectors`
  - Display a `GridView` containing `ColorSwatch` components, sorted alphabetically by color name
  - Add new actions in `application` reducer for selecting, adding, and removing shared colors (for selection, see the `selectLayer` action. we can pass a single ID, an array of ID, or undefined to select)
  - We can add a new top-level array in `application` state for `selectedSwatchIds`
- Create a new inspector, maybe `SwatchesInspector`, to allow editing colors
  - E.g. in `Workspace.tsx`, render something like `<RightSidebar><SwatchesInspector /></RightSidebar>`
  - This probably *shouldn't* go within the existing `Inspector`, which should maybe be renamed to `LayerInspector`
  - For color picker usage, see `ColorInputField.tsx`
  - Also allow editing the name of the color (you can use the `InputField` component)
- Create a new toolbar, maybe `SwatchesToolbar`, in `containers` and use it in `Workspace` instead of the existing `Toolbar`
  - Add a button in the toolbar to add a new color
  - There should be a `<PlusIcon />` from the radix icon set you can use

## UX

This is what Sketch does, but it doesn't need to look exactly the same. We can use different colors, sizes, etc, to make it look nice in our UI.

https://user-images.githubusercontent.com/1198882/110185686-dcb8e200-7dc7-11eb-89bd-557d15ce1d2c.mov

And here's adding a color.

https://user-images.githubusercontent.com/1198882/110185895-75e7f880-7dc8-11eb-89cd-b6fb9f2be33c.mov

milg15:
There is a lot of room for improvement 😅  But I think it resolves #5 for the most part
image


packages/noya-state/src/selectors.ts
```diff
@@ -26,6 +26,10 @@ export const getCurrentPageIndex = (state: ApplicationState) => {
   return pageIndex;
 };
 
+export const getSharedSwatches = (state: ApplicationState) => {
+  return state.sketch.document.sharedSwatches ?? [];
```
dabbott, line 30: From our call, we can probably return a `Swatch[]` instead of `SwatchContainer`:

```ts
export const getSharedSwatches = (
  state: ApplicationState,
): Sketch.Swatch[] => {
  return (
    state.sketch.document.sharedSwatches?.objects ?? []
  );
};
```

packages/noya-state/src/selectors.ts
```diff
@@ -125,6 +129,27 @@ export const getSelectedLayersWithFixedRadius = (
   ).filter((layer): layer is Sketch.Rectangle => layer._class === 'rectangle');
 };
 
+export const getSelectedSwatches = (
```
dabbott, line 132: If this isn't used we should remove it, cause I'm not sure if there are other kinds of swatches besides colors. I think the other types are in a different place in the file format

packages/noya-state/src/selectors.ts
```diff
@@ -125,6 +129,27 @@ export const getSelectedLayersWithFixedRadius = (
...
+  const swatches = sharedSwatches.objects;
+  return swatches.filter((swatch) =>
+      state.selectedSwatchIds.includes(swatch.do_objectID),
+    )
+};
+
+export const getSelectedColorSwatches = (
+  state: ApplicationState,
+): Sketch.Swatch[] => {
+  const sharedSwatches = getSharedSwatches(state) as Sketch.SwatchContainer;
```
dabbott, line 145: Remove `as`

packages/noya-state/src/reducers/interaction.ts
```diff
@@ -52,7 +52,9 @@ export type InteractionAction =
   | [type: 'updateMoving', point: Point]
   | [type: 'updateScaling', point: Point]
   | [type: 'updatePanning', point: Point]
-  | [type: 'enablePanMode'];
+  | [type: 'enablePanMode']
+  | [type: 'modifyColor']
```
dabbott, line 56: Remove these since they're unused

packages/noya-state/src/reducers/application.ts
```diff
@@ -874,6 +1002,17 @@ function accessPageLayers(
   });
 }
 
+function createColorSketch(
```
dabbott, line 1005: Unused?

packages/noya-state/src/reducers/application.ts
```diff
@@ -113,6 +119,19 @@ export type Action =
   | [type: 'setOpacity', amount: number, mode?: SetNumberMode]
   | [type: 'setFixedRadius', amount: number, mode?: SetNumberMode]
   | [type: `set${StyleElementType}Color`, index: number, value: Sketch.Color]
+  | [
+      type: 'setSwatchColor', 
+      swatchId: string | string[] | undefined,
```
dabbott, line 124: `undefined` isn't a valid option here, since we need at least 1 swatch selected to change its color

packages/noya-state/src/reducers/application.ts
```diff
@@ -113,6 +119,19 @@ export type Action =
...
+      color: Sketch.Color
+    ]
+  | [
+      type: 'setSwatchName', 
+      swatchId: string | string[] | undefined,
```
dabbott, line 129: Same as above, `undefined` isn't a valid option here

packages/noya-state/src/reducers/application.ts
```diff
@@ -853,6 +871,116 @@ export function reducer(
...
+          "blue": 0
+        }
+        const swatch: Sketch.Swatch = {
+          "_class": "swatch",
+          "do_objectID": uuid(),
+          "name": "Undefined",
+          "value": swatchColor,
+        }
+
+        sharedSwatches?.objects.push(swatch);
```
dabbott, line 892: We need to remove the `?` so that this still works if there are no swatches yet. Maybe we can use the `getSharedSwatches` selector? (once we update it)

packages/noya-state/src/reducers/application.ts
```diff
@@ -853,6 +871,116 @@ export function reducer(
...
+            });
+            return;
+          case 'replace':
+            state.selectedSwatchIds = [...ids];
+            return;
+        }
+      });
+    }
+    case 'setSwatchColor': {
+      const [, swatchId, color] = action;
```
dabbott, line 921: We should always "normalize" the swatchId to an array here so the rest of the code is simpler.

E.g. `const swatchIds = typeof swatchId === 'string' ? [swatchId] : swatchId`;

packages/noya-state/src/reducers/application.ts
```diff
@@ -853,6 +871,116 @@ export function reducer(
...
+            return swatch.do_objectID === swatchId;
+          }
+          else {
+            const ids = swatchId as string[];
+            return ids.includes(swatch.do_objectID);
+          }
+        };
+
+        index = sharedSwatches.objects.findIndex(isTheSwatch);
+        while (index != -1) {
```
dabbott, line 946: This can probably be a `forEach`

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -0,0 +1,16 @@
+import {
```
dabbott, line 1: From our call, example GridView usage might look like:

```ts
import * as GridView from './GridView'

function MyGrid() {
  const items = []

  return (
    <GridView.Root onClick={() => {
      dispatch('selectSwatch', undefined)
    }}>
      {items.map(item => {
        <GridView.Item title={""} subtitle={""} onClick={(event) => {
          event.stopPropagation()
          event.preventDefault()

          dispatch('selectSwatch', item.do_objectID)
        }}>{children}</GridView.Item>
      })}
    </GridView.Root>
  )
}
```

So we should export a `Root` and an `Item` component from this file.

packages/app/src/containers/Workspace.tsx
```diff
@@ -47,6 +51,7 @@ const MainView = styled.main(({ theme }) => ({
 const ContentArea = styled.div(({ theme }) => ({
   flex: '1',
   display: 'flex',
+  overflow: 'hidden'
```
dabbott, line 54: I think we can add a wrapper to the element with `display: grid` that has `overflowY: auto` to fix this. We do that in `GridView.tsx` so that other components don't need to think about it

packages/app/src/containers/SwatchesToolbar.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export default function Toolbar() {
+  const [state, dispatch] = useApplicationState();
+  const itemSeparatorSize = useTheme().sizes.toolbar.itemSeparator;
+  const interactionType = state.interactionState.type;
+
+  return useMemo(
+    () => (
+      <Container>
+        <div>
```
dabbott, line 27: We can remove the title, we don't need it yet, and it may go somewhere else

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    let swatches: Array<Sketch.Swatch> = []
+    if (sharedSwatches){
+        sharedSwatches = sharedSwatches as Sketch.SwatchContainer
+        swatches = sharedSwatches.objects;
+    }
+
+    const handleMouseDown = useCallback(
+      (event: React.PointerEvent, swatch?: Sketch.Swatch) => {  
+        switch (state.interactionState.type) {
```
dabbott, line 40: We can use `event.stopPropagation()` and `event.preventDefault()` to fix text being selection when shift-clicking.

Here's an example of how we handle the event in list views:

https://github.com/noya-app/noya/blob/main/packages/noya-designsystem/src/components/ListView.tsx#L127-L134

And how to handle shift select in the layer list (although yours will be simpler, since we don't need to handle meta key or handle selecting ranges of items):

https://github.com/noya-app/noya/blob/main/packages/app/src/containers/LayerList.tsx#L99-L126

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -0,0 +1,302 @@
...
+  const swatch: Sketch.Swatch = swatches[0];
+
+  const color = swatch.value as Sketch.Color;
+  const rgbObj = {
+    "r": Math.round(color.red   * 255),
+    "g": Math.round(color.green * 255),
+    "b": Math.round(color.blue  * 255)
+  }
+
+  function rgb2hex(rgb: string){
```
dabbott, line 228: Is there an existing function we can use from https://github.com/noya-app/noya/blob/6af9eb378e9a26cdb5d241184864249991819c25/packages/noya-colorpicker/src/utils/convert.ts?

packages/app/src/containers/ColorSwatchGrid.tsx
```diff
@@ -0,0 +1,38 @@
+
+import styled from 'styled-components';
+
+const ScrollContainer = styled.div`
```
dabbott, line 4: Handling custom scroll can look nice, but I find it more accessible to use the built-in scrollbar. It also handles theming and different operating systems automatically

packages/app/src/App.tsx
```diff
@@ -15,7 +15,7 @@ import { useResource } from './hooks/useResource';
 import { PromiseState } from './utils/PromiseState';
 
 export default function App() {
-  const sketchFile = useResource<ArrayBuffer>('/Demo.sketch', 'arrayBuffer');
+  const sketchFile = useResource<ArrayBuffer>('/ColorAsset.sketch', 'arrayBuffer');
```
dabbott, line 18: Undo before we merge

packages/app/src/components/swatches/ColorSwatch.tsx
```diff
@@ -0,0 +1,77 @@
+import { RgbaColor } from 'noya-colorpicker/src/types';
+import Sketch from '@sketch-hq/sketch-file-format-ts';
+import styled from 'styled-components';
+import { memo } from 'react';
+
+
+const ColorSwatchContainer = styled.div(({ selected }: PropsSelected) => ({
```
dabbott, line 7: We should use theme colors here to support light/dark mode

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -0,0 +1,91 @@
...
+  getSharedSwatches
+} from 'noya-state/src/selectors';
+import styled from 'styled-components';
+
+/**
+ * TODO: Multiselect and change color 
+ * TODO: Clean code
+ */
+
+const SwatchContainer = styled.div`
```
dabbott, line 18: Use styled.div({}) (object notation) instead of styled.div`` here for consistency

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -0,0 +1,91 @@
...
+  height: 175px,
+  &:hover {
```
dabbott, line 20: This would be `"&:hover": { ... }`, although I think in this case we don't need the `:hover`, since the cursor property only matters when the element is hovered anyway

dabbott (changes requested):
Ok! I tried to add a note on everything we talked about. 

I'm not sure about the best way to rearrange things in the color swatch files yet, but I think we can revisit that after you update the PR

packages/noya-state/src/selectors.ts
milg15: I updated this!

packages/noya-state/src/selectors.ts
milg15: No more `as` yay!

packages/noya-state/src/selectors.ts
milg15: Deleted the extra selector

packages/noya-state/src/reducers/interaction.ts
milg15: I can confirm that it wasn't used. I deleted it.

packages/noya-state/src/reducers/application.ts
milg15: Yes, it was unused. I deleted it

packages/noya-state/src/reducers/application.ts
milg15: it only takes`string | string[]` now

packages/noya-state/src/reducers/application.ts
milg15: I modified it to create it in case it doesn't exist.

packages/noya-state/src/reducers/application.ts
milg15: Normalized

packages/noya-state/src/reducers/application.ts
milg15: it's now a `forEach`

packages/noya-designsystem/src/components/GridView.tsx
milg15: It  uses the `GridView` now instead of the awkward events. Although the preventDefault it's not really working. I'm not sure if I am doing something wrong.

packages/app/src/containers/Workspace.tsx
milg15: I changed to use the container you suggested. It's all in `GridView.tsx` now.

packages/app/src/containers/SwatchesToolbar.tsx
milg15: Oh! I left it for now 😅

packages/app/src/containers/ComponentsWindow.tsx
milg15: Click events are done in a more Reactive way 🥳

packages/app/src/containers/ColorSwatchesInspector.tsx
milg15: Now using the existing function

packages/app/src/containers/ColorSwatchGrid.tsx
milg15: using the built-in scrollbar.

packages/app/src/App.tsx
milg15: Changed it in last commit

packages/app/src/components/swatches/ColorSwatch.tsx
milg15: I think this can be updated to use more of the theme colors.  Maybe create a theme for this section?

dabbott: Sure - if you can get it pretty close with existing colors I can also tweak it a bit after we merge it too

packages/app/src/components/inspector/ColorSelectRow.tsx
```diff
@@ -0,0 +1,191 @@
...
+      onNudgeOpacity(amount / 100);
+    },
+    [onNudgeOpacity],
+  );
+
+  return (
+    <Column>
+      <InputField.Root id={'colorName'}>
+        <InputField.Input
+          value={name === 'Multiple' ? '' : name}
```
dabbott, line 161: In the case of multiple selections: let's set the `value` to empty string, and use `"Multiple"` as the `placeholder`

Perhaps change the name prop to be `string | undefined`? Or maybe there's a better way

packages/app/src/components/inspector/ColorSelectRow.tsx
```diff
@@ -0,0 +1,191 @@
...
+}));
+
+const Column = styled.div(({ theme }) => ({
+  flex: '1',
+  display: 'flex',
+  flexDirection: 'column',
+}));
+
+const FullContent = styled.div(({ theme }) => ({
+  width: '240px',
```
dabbott, line 37: Let's remove the width here, since it was only needed for the popup color picker. This one can fill its parent. Also, padding in this case I think should be `10px 0`, since we don't need more horizontal padding.

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const name =
+    selectedSwatches.length > 1 &&
+    (firstSwatch.name === 'New Color Variable' || !sameName())
+      ? 'Multiple'
+      : firstSwatch.name;
+
+  const ids = state.selectedSwatchIds;
+  return (
+    <Container>
+      <div style={{ display: 'flex', alignItems: 'center' }}>
```
dabbott, line 55: Can we remove this `div`? I don't think it does anything here

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <Title>{'Name'}</Title>
+        <Spacer.Horizontal size={12} />
+      </div>
+      <ColorSelectRow
+        id={`fill-${'color'}`}
```
dabbott, line 60: The `id` was previously dynamic so it would work in lists, but here we only have a single color picker, so it can be just `"color-swatch"` or something

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -0,0 +1,98 @@
...
+        }
+        onChangeColor={(value) => dispatch('setSwatchColor', ids, value)}
+        onInputChange={(value) => dispatch('setSwatchName', ids, value)}
+      />
+    </Container>
+  );
+});
+
+export default memo(function SwatchesInspector() {
+  const currentTab = useSelector(Selectors.getCurrentTab);
```
dabbott, line 76: I don't think we need to check the `currentTab` here, since as it's written now, this component should still behave correctly even if theoretically used on a different tab

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <GridView.Item
+          key={index}
+          title={item.name}
+          subtitle={createSubtitle(item)}
+          selected={state.selectedSwatchIds.includes(item.do_objectID)}
+          onClick={(event) => {
+            event.stopPropagation();
+            event.preventDefault();
+
+            console.log(event);
```
dabbott, line 40: Remove log

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useApplicationState } from '../contexts/ApplicationStateContext';
+import { getSharedSwatches } from 'noya-state/src/selectors';
+import { rgbaToHex } from 'noya-colorpicker';
+
+export default memo(function ComponentsWindow() {
+  const [state, dispatch] = useApplicationState();
+
+  const items = getSharedSwatches(state);
+
+  const createSubtitle = (item: Sketch.Swatch) => {
```
dabbott, line 14: Since `item` has type `Sketch.Swatch`, it can never have a `_class` other than `'swatch'`, so I don't think we need to check for it.

In terms of coding style, I would do this within the `items.map`

E.g.

```ts
items.map((item, index) => {
  const color = {
    a: item.value.alpha,
    r: Math.round(item.value.red * 255),
    g: Math.round(item.value.green * 255),
    b: Math.round(item.value.blue * 255),
  };

  return (
    <GridView.Item
      key={index}
      title={item.name}
      subtitle={`${rgbaToHex(color)} - ${Math.round(color.a * 100)}%`}
      selected={state.selectedSwatchIds.includes(item.do_objectID)}
      onClick={(event) => {
        event.stopPropagation();
        event.preventDefault();

        dispatch(
          'selectSwatch',
          item.do_objectID,
          event.shiftKey ? 'intersection' : 'replace',
        );
      }}
    >
      <ColorSwatch value={item.value} />
    </GridView.Item>
  )
})
```

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -0,0 +1,98 @@
...
+    r: Math.round(firstSwatch.value.red * 255),
+    g: Math.round(firstSwatch.value.green * 255),
+    b: Math.round(firstSwatch.value.blue * 255),
+    a: firstSwatch.value.alpha,
+  };
+
+  let sameName = () =>
+    selectedSwatches.every((v) => v.name === firstSwatch.name);
+
+  const name =
```
dabbott, line 46: I would use `undefined` to represent multiple different names, and handle that specially within the `ColorSelectRow`

```ts
const name = selectedSwatches.length > 1 && !sameName() ? undefined : firstSwatch.name
```

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -0,0 +1,98 @@
...
+    (firstSwatch.name === 'New Color Variable' || !sameName())
+      ? 'Multiple'
+      : firstSwatch.name;
+
+  const ids = state.selectedSwatchIds;
+  return (
+    <Container>
+      <div style={{ display: 'flex', alignItems: 'center' }}>
+        <Title>{'Name'}</Title>
+        <Spacer.Horizontal size={12} />
```
dabbott, line 57: Should it be `Spacer.Vertical`?

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -0,0 +1,98 @@
...
+      </div>
+      <ColorSelectRow
+        id={`fill-${'color'}`}
+        color={selectedSwatches[0].value}
+        name={name}
+        hexValue={rgbaToHex(firstColor)}
+        onChangeOpacity={(value) => dispatch('setSwatchOpacity', ids, value)}
```
dabbott, line 64: We should wrap all of these callbacks in `useCallback` (we should probably do this for most callbacks in other files too)

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <GridView.Root onClick={() => dispatch('selectSwatch', undefined)}>
+      {items.map((item, index) => (
+        <GridView.Item
+          key={index}
+          title={item.name}
+          subtitle={createSubtitle(item)}
+          selected={state.selectedSwatchIds.includes(item.do_objectID)}
+          onClick={(event) => {
+            event.stopPropagation();
```
dabbott, line 37: Can we move `event.stopPropagation()` and `event.preventDefault()` into `GridView.tsx`? I think we will want these for every click event

packages/app/src/containers/SwatchesToolbar.tsx
```diff
@@ -0,0 +1,48 @@
...
+}));
+
+export default function Toolbar() {
+  const [, dispatch] = useApplicationState();
+  const itemSeparatorSize = useTheme().sizes.toolbar.itemSeparator;
+
+  return useMemo(
+    () => (
+      <Container>
+        <Title>
```
dabbott, line 30: Remove `Title` and spans inside, we don't need this UI element yet. The first horizontal spacer can have size 8

packages/noya-designsystem/src/components/ColorInputField.tsx
```diff
@@ -50,6 +50,7 @@ interface Props {
   onChange: (color: FileFormat.Color) => void;
 }
 
+
```
dabbott, line 53: If you format with prettier, this should go away

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -0,0 +1,103 @@
+import { memo, useCallback, ReactNode } from 'react';
+import styled from 'styled-components';
+
+const Grid = styled.div(({ theme }) => ({
+  flex: 1,
+  color: theme.colors.text,
+  display: 'grid',
+  padding: '14px',
+  gridTemplateColumns: 'repeat(auto-fill, 200px)',
+  gridTemplateRows: 'repeat(auto-fill, 180px)', // Makes it go
```
dabbott, line 10: I think you're right that we do need this, we can remove the comment

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -0,0 +1,103 @@
...
+  justifyContent: 'space-between',
+  gap: '20px',
+  overflowY: 'scroll',
```
dabbott, line 13: We can use `overflowY: 'auto'` here - this will conditionally show a scrollbar only if needed. `'scroll'` shows a scrollbar all the time, which is usually not what we want

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -0,0 +1,103 @@
...
+const Container = styled.div(({ theme }) => ({
+  flex: '1',
+  display: 'flex',
+  flexDirection: 'column',
+}));
+
+const ItemContainer = styled.div<{ selected: boolean }>(
+  ({ theme, selected }) => ({
+    width: '200px',
+    height: '130px',
```
dabbott, line 32: Instead of hardcoding the height here, let's make `SwatchContainer` a flex container with `display: 'flex', flexDirection: 'column'`. Then this `ItemContainer` can be `flex: 1` to fill the available space.

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -0,0 +1,103 @@
...
+    justifyContent: 'center',
+    display: 'flex',
+    borderRadius: '12px',
+    cursor: 'pointer',
+    border: `2px ${selected ? 'solid' : 'none'} rgb(132,63,255)`,
+  }),
+);
+
+const SwatchContainer = styled.div(({ theme }) => ({
+  heigth: '180px',
```
dabbott, line 44: Typo on "height", but I think we can remove this anyway and use flex 😂

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -0,0 +1,103 @@
...
+}));
+
+const ItemTitle = styled.span(({ theme }) => ({
+  color: theme.colors.textDecorativeLight,
+  margin: '0px',
```
dabbott, line 49: Does margin 0 do anything here?

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -0,0 +1,103 @@
...
+}));
+
+const ItemDescription = styled.div(({ theme }) => ({
+  color: theme.colors.text,
+  margin: '0px',
```
dabbott, line 54: Also here?

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -0,0 +1,103 @@
...
+  );
+}
+
+interface GridViewRootProps {
+  children?: ReactNode;
+  onClick?: () => void;
+}
+
+function GridViewRoot({ children, onClick }: GridViewRootProps) {
+  const handleClick = useCallback(
```
dabbott, line 83: This is good! Let's do this for items also, to encapsulate the `event.stopPropagation()` and `event.preventDefault()`

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -0,0 +1,53 @@
...
+            event.preventDefault();
```
dabbott, line 38: Maybe `preventDefault` doesn't prevent selection... I think we might still need `user-select: none` on the text

packages/noya-state/src/reducers/interaction.ts
```diff
@@ -52,7 +52,7 @@ export type InteractionAction =
   | [type: 'updateMoving', point: Point]
   | [type: 'updateScaling', point: Point]
   | [type: 'updatePanning', point: Point]
-  | [type: 'enablePanMode'];
+  | [type: 'enablePanMode']
```
dabbott, line 55: Formatting with prettier should undo this

packages/noya-state/src/selectors.ts
```diff
@@ -125,6 +133,16 @@ export const getSelectedLayersWithFixedRadius = (
...
 };
 
+export const getSelectedColorSwatches = (
+  state: ApplicationState,
+): Sketch.Swatch[] => {
+  const sharedSwatches = getSharedSwatches(state);
+
+  return sharedSwatches.filter((swatch) =>
+      state.selectedSwatchIds.includes(swatch.do_objectID),
+    ).filter(swatch => swatch.value._class === 'color');
```
dabbott, line 143: I don't think we need to filter on `_class`. The Sketch types are weird and I think the other shared styles aren't in the swatch object

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -0,0 +1,53 @@
...
+      };
+
+      return `${rgbaToHex(color)} - ${Math.round(color.a * 100)}%`;
+    }
+    return '';
+  };
+
+  return (
+    <GridView.Root onClick={() => dispatch('selectSwatch', undefined)}>
+      {items.map((item, index) => (
```
dabbott, line 30: Sketch sorts color swatches by name when presenting them in the grid - let's do the same

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <Container>
+      <div style={{ display: 'flex', alignItems: 'center' }}>
+        <Title>{'Name'}</Title>
+        <Spacer.Horizontal size={12} />
+      </div>
+      <ColorSelectRow
+        id={`fill-${'color'}`}
+        color={selectedSwatches[0].value}
+        name={name}
+        hexValue={rgbaToHex(firstColor)}
```
dabbott, line 63: The "#" is redundant here, since it's already in the input label. We could do `rgbaToHex(firstColor).slice(1)` to remove it

Screen Shot 2021-03-15 at 2 48 20 PM

dabbott (changes requested):
Nice! I think we've taken care of all the big things, now it's just a lot of little stuff.

After this PR I will add a git precommit hook for automatically applying prettier 😅

packages/app/src/components/swatches/ColorSwatch.tsx
milg15: Okey! I think overall used it in the background and text. The exact ones I am no really sure

packages/noya-designsystem/src/components/GridView.tsx
milg15: Without adding the height. It looks like this.
image
Is there any other way? To not have the height hardcoded. Here is the code:

```javascript
const SwatchContainer = styled.div(({ theme }) => ({
  display: 'flex',
  flexDirection: 'column',
  flex: '1',
}));

const ItemContainer = styled.div<{ selected: boolean }>(
  ({ theme, selected }) => ({
    display: 'flex',
    flex: '1',
    (...)
  }),
);

```

packages/noya-designsystem/src/components/GridView.tsx
milg15: it did when it was a p, but not needed anymore with the new layout. You're right

packages/noya-designsystem/src/components/GridView.tsx
dabbott: I think we may want to use a fixed height for rows. I have never used this property before 😂  but it seems to work:

```ts
const Grid = styled.div(({ theme }) => ({
  // gridTemplateRows: 'repeat(auto-fill, 180px)'
  gridAutoRows: '180px',
  // ...
}));
```

dabbott: Minor thing: we probably don't need `flex: 1` on the SwatchContainer, since its parent is `display: grid` so it would have no effect

milg15: Oh I see, that way the height is not in the item itself. Great!

packages/app/src/containers/ComponentsWindow.tsx
milg15: I have a question about this one. I sorted them but when I add a new one should I place it at the end? and if so if I add two new ones should the previous one stay at the bottom or should it move with the name "New Color Variable" where it belongs ?

dabbott: I think new ones should be sorted too, rather than placed at the end. 

Ideally a new color would also scroll into view if needed, and the name input field would be focused automatically, but we can do that in the future

milg15: I understand I'll leave it as it is for now. It's sorted

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -9,45 +8,49 @@ import { rgbaToHex } from 'noya-colorpicker';
...
+          >
+            <ColorSwatch value={item.value} />
+          </GridView.Item>
+        );
+      })}
     </GridView.Root>
   );
 });
+
+/**
```
dabbott, line 50: Remove comment

dabbott (changes requested):
This is very close! Most of my remaining feedback is just nitpicks

packages/app/src/containers/ColorSwatchesInspector.tsx
```diff
@@ -46,35 +46,41 @@ const ColorPickerInspector = memo(function ColorPickerInspector() {
   const name =
     selectedSwatches.length > 1 &&
     (firstSwatch.name === 'New Color Variable' || !sameName())
-      ? 'Multiple'
+      ? undefined
```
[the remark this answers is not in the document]
milg15: If all the selected colors are New I want to show the `Multiple` placeholder anyway. If not, when they choose multiple new colors it will show "New Color Variable" in the textbox. Maybe it's okey if it looks like this

dabbott (approved):

Issue #8, opened 2021-03-18
Add color swatches to color picker


dabbott:
## Feature

When picking a color for borders, fills, and shadows, we should show the document's color swatches in the color picker popup.

## Possible implementation

- The main file here is `ColorInputFieldWithPicker.tsx`. 
- We should show the document's `sharedSwatches` and allow choosing one to set the color
  - The `Sketch.Color` type has an optional `swatchID` field - we can use this field to determine if we should show an item as selected in the grid/list. This also means we should set the `swatchID` when the user picks a color swatch.
- We should support 2 layouts: a grid and a list. We may want to make a separate file for each of these components.
  - For the **list** layout, we should use our `ListView` component. It's OK if it the spacing looks weird - I can fix that later - let's just focus on functionality for now
  - I think the **grid** layout can be a new component (probably using CSS grid again). It's pretty different from our `GridView`, so I don't think we need to reuse that.
- We shouldn't need to handle scrolling specially - the whole popup should scroll if the list gets long, and I think this already works correctly.
- Ignore the `"Document colors"` dropdown for now, since we don't have any other kinds of colors yet

## UX

Here's a mockup of the two views:

| Grid | List |
| --- | --- |
| ![swatch-grid](https://user-images.githubusercontent.com/1198882/111700577-d169ab80-87f6-11eb-992c-ac16550e8050.png) | ![swatch-list](https://user-images.githubusercontent.com/1198882/111700555-cadb3400-87f6-11eb-965c-7a346e3544e6.png) |

> Note that these aren't totally accurate, since they show a selected swatch, but the color picker itself shows a different color. The color picker on the left should be orange and the color picker on the right should be red, but instead they're both blue

milg15:
Should I show all the colors? or limit it to a few in the grid view? in case there were a lot

dabbott:
Let's show all of them. I added an `overflowY: auto` to the popup already so it should look reasonable. Maybe in the future we can add a more sophisticated version, like with search.

milg15:
Do I create a new branch for this? Or do I keep it in the color-swatches branch?

PR #9, opened 2021-03-21
design and action completed (#8)

milg15:
I think this could be move to another file but you can validate me first if the implementation it's good to go 😄

### DIFF
--- a/packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
+++ b/packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
@@ -78,13 +78,13 @@ interface Props {
 interface SwatchesProps {
   selectedSwatchId: string;
   swatches: Sketch.Swatch[];
-  onClick: (color: Sketch.Color) => void;
+  onSelectSwatch: (color: Sketch.Color) => void;
 }
 
 const SwatchesList = memo(function SwatchesList({
   selectedSwatchId,
   swatches,
-  onClick,
+  onSelectSwatch,
 }: SwatchesProps) {
   return (
     <ListView.Root>
@@ -97,7 +97,7 @@ const SwatchesList = memo(function SwatchesList({
             key={item.do_objectID}
             selected={selectedSwatchId === item.do_objectID}
             onClick={() => {
-              onClick({
+              onSelectSwatch({
                 ...item.value,
                 swatchID: item.do_objectID,
               });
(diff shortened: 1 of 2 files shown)


packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
```diff
@@ -52,6 +76,72 @@ interface Props {
...
+              rgbaToHsva(sketchColorToRgba(item.value)),
+            );
+
+            return (
+              <Square
+                key={item.do_objectID}
+                color={colorString}
+                selected={selected === item.do_objectID}
+                onClick={() => {
+                  onClick(item.value);
```
dabbott, line 108: Ok - I don't think I explained this part very well... colors have a property called `swatchID` which we should set when a swatch is selected. For example:

```ts
onClick({
  ...item.value,
  swatchID: item.do_objectID,
});
```

We can then use this to show selection. So for example, `GridProps` could contain a `selectedSwatchId` prop, and then 

```tsx
<Square 
  selected={selectedSwatchId === item.do_objectID} 
  // ... 
/>
```

Then in the `ColorInputFieldWithPicker` we'll render using 

```tsx
<SwatchesSelector
  selectedSwatchId={value.swatchID} 
  // ...
/>
```

Hopefully that's clear - if not let me know

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
```diff
@@ -52,6 +76,72 @@ interface Props {
   onChange: (color: Sketch.Color) => void;
 }
 
+interface GridProps {
+  id?: string;
```
dabbott, line 80: `selectedSwatchId` maybe?

dabbott (changes requested):
Looks good! It's pretty concise. I'm thinking we can leave it all in this file for now.

packages/app/src/components/swatches/ColorSwatch.tsx
```diff
@@ -17,10 +16,7 @@ interface Props {
 }
 
 export default memo(function ColorSwatch({ value }: Props) {
-  const colorString = useMemo(
-    () => hsvaToRgbaString(rgbaToHsva(sketchColorToRgba(value))),
-    [value],
-  );
+  const colorString = useMemo(() => sketchColorToRgbaString(value), [value]);
```
milg15, line 19: I changed it so it uses the proper function

dabbott (approved):

dabbott:
Looks great! I'll go ahead and merge it 🚀

milg15:
awesome 🎉

PR #11, opened 2021-03-30
Shared styles

### LINKED ISSUE #10: Add shared styles to components tab
## Feature

Just like shared color swatches, we should allow the user to create and edit shared layer styles.

What's included in this task: 
- Implementing the tabs at the top to switch between shared colors/layer styles
- Adding new shared layer styles via the "+" button
- The inspector on the right, to allow editing shared layer styles

For example, here's what this looks like in Sketch:

Screen Shot 2021-03-24 at 1 59 40 PM

Screen Shot 2021-03-24 at 1 59 46 PM

## File format notes

Here's what I've learned from looking at `.sketch` file JSON and the TypeScript types.

A layer style has the type `Style`. In our code this will usually be `Sketch.Style`.

Screen Shot 2021-03-24 at 2 04 16 PM

All layers use this style: Rectangles, Ovals, Bitmaps, Groups, Artboards, etc. However, some properties are specific to certain types of layers. For example, `textStyle` only applies to Text layers, and `colorControls` only applies to Bitmap layers.

Shared styles for a document are stored in `state.sketch.document.layerStyles`.

Screen Shot 2021-03-24 at 2 10 12 PM

**The following won't be part of this issue, but may be useful to know**

A **style** object has a `do_objectID` just like most other Sketch types. When applying a shared style to a layer, we should set the `sharedStyleID` **on the layer** to the `do_objectID` of the style. We should additionally **clone** the shared style but assign a new id (similar to e.g. https://github.com/noya-app/noya/blob/40e4014ef9e79579a1b5359458cbbd588d4a92e6/packages/noya-state/src/reducers/application.ts#L184)

Screen Shot 2021-03-24 at 2 15 20 PM

The `sharedStyleID` maintains the link between the shared style and the layer it's applied to. The cloned layer `style` is actually (as far as I can tell) not linked to the `sharedStyleID` in any way and can be changed independently! We will handle updating layer styles manually, by traversing the layer hierarchy, and finding which styles need to update based on the `sharedStyleID`.

## Possible implementation

- I've added a file `SharedStyles.sketch` that already has 2 shared styles which you can use for testing
- The first step is to implement that tabs at the top of the components window. I've already add a few of the UI components, but we need to store something in the `state` and add a new action in `application.ts` to handle switching tabs
  Screen_Shot_2021-03-24_at_2_29_54_PM
- The behavior of the `+` button will depend on which tab is selected: if we're on the colors tab, it will add a new swatch. If we're on the layer styles tab, it will add a new layer style. We can determine this via the current tab (in `state`)
- The inspector UI for this should already exist. However, we will likely want to modify the existing inspector components to work with both layers styles and shared styles. 
  - I'm not sure exactly the best way to do this. 
  - My suggestion would be: make a new component `StyleInspector`, that takes an array of ids, `styleIds: string[]` as a prop. 
  - We can move most of the components within `Inspector` into `StyleInspector`: `<OpacityInspector />`, `<FillInspector />`, `<BorderInspector />`, `<ShadowInspector />`. These should all _also_ take `styleIds` as props. The `Inspector` can then render a `StyleInspector`. **This may be somewhat challenging, so we could also do it together / pair on it if that would be helpful**
  - You're welcome to try a different approach, or to copy & paste to start, and we can refactor after like we did for color inspector
  - If there are properties that you see in the screenshots above that don't currently exist in the inspector UI (such as inner shadows, blurs) you don't need to add them.
- Display a preview of the color style using HTML & CSS. We should show a square with a background color, a border, and a shadow. Anything fancy or difficult (e.g. radial gradients) we can do another time. This should be similar to our color swatch grid.
  Screen Shot 2021-03-24 at 2 43 15 PM
- **In the future** we will add a picker for selecting a shared style from a layer (similar to using a shared color swatch in a fill). And when a shared layer style changes, we will also need to update any layers that currently use that style. However, we don't need to handle these cases as part of this issue.

milg15:
The border preview is a work in progress and the code needs some refactoring but everything else should work a-okey.

closes #10 
image


packages/app/src/components/component/LayerStyle.tsx
```diff
@@ -0,0 +1,105 @@
...
+    width: '65px',
+    border: border,
+    borderRadius: '10px',
+    background: background,
+    boxShadow: shadow,
+    opacity: opacity,
+  }),
+);
+
+// TODO: Handle multiple borders in preview.
```
dabbott, line 29: This is fine for now. We'll improve this as a separate task.

packages/app/src/components/inspector/FillRow.tsx
```diff
@@ -79,7 +80,10 @@ export default memo(function FillRow({
         />
         <Spacer.Horizontal size={8} />
         <InputField.Root id={hexInputId} labelPosition="start">
-          <InputField.Input value={'FFFFFF'} onSubmit={() => {}} />
+          <InputField.Input
+            value={`${sketchColorToHex(color).replace('#', '').toUpperCase()}`}
```
dabbott, line 84: We can probably move `toUpperCase()` directly into the `sketchColorToHex` function. I think we always want uppercase in this app

packages/noya-state/src/reducers/application.ts
```diff
@@ -459,263 +517,38 @@ export function reducer(
...
-      const [, index, amount, mode = 'replace'] = action;
+    case 'setShadowColor':
+    case 'setBorderWidth':
+    case 'setFillOpacity':
+    case 'setOpacity':
+    case 'setShadowX':
+    case 'setShadowY':
+    case 'setShadowBlur':
+    case 'setBorderPosition':
+    case 'setShadowSpread': {
```
dabbott, line 543: These cases now delegate to the `styleReducer` function. We should do the same with the version of these for shared layer styles

dabbott (changes requested):
This is looking good! It's helpful to see it all written out like this, so we can figure out the best way to consolidate.

I added a `style.ts` file containing a `styleReducer` function directly to your branch. We can hopefully use this to share code between actions that affect the selected layer + actions that affect shared layer styles. I was able to combine a lot of the switch cases into one, which I think we should be able to do for many of the actions that you added (any action that modifies a style). Hopefully this will shorten the code a bit.

I think this PR will require another round of feedback/updates around consolidating inspector components, but let's start with these comments and then I'll take another look.

packages/noya-state/src/reducers/application.ts
```diff
@@ -1099,26 +782,92 @@ export function reducer(
...
+              state.selectedLayerStyleIds.splice(selectedIndex, 1);
+            });
+            return;
+          case 'replace':
+            state.selectedLayerStyleIds = [...ids];
+            return;
+        }
+      });
+    }
+    case 'setComponentName': {
```
dabbott, line 855: While this does result in shorter code, I actually prefer having separate actions for renaming different types. However, if it simplifies things, you could try combining them into one switch case, e.g.:

```ts
case 'setSwatchName':
case 'setSharedStyleName':
case 'setSharedTextStyleName':
case 'setSymbolName': {
  /* one implementation */
}
```

I don't mind them being separate and duplicated though, if that's more clear. We have this duplication already for border, style, shadows, etc, and I don't think it's a bad thing.

packages/app/src/containers/NameInspector.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { memo, useCallback } from 'react';
+import * as InspectorPrimitives from '../components/inspector/InspectorPrimitives';
+import { useApplicationState } from '../contexts/ApplicationStateContext';
+
+interface Props {
+  names: string[];
+  ids: string[];
+}
+
+export default memo(function NameInspector({ names, ids }: Props) {
```
dabbott, line 11: I think I preferred how you had it before, where this component didn't rely on `useApplicationState` at all.

I'd like the inspectors for swatches, layer styles, text styles, and symbols to all be as self-contained as possible, even if it results in some duplicated `<NameInspector />`s, so that they're more reusable.

Also, we should try not to mix `id`s for different kinds of objects. Here an `id` might refer to a swatch or a style, but we have no way to know. I see this is a weakness of TypeScript for not giving us a good way to enforce different id types.

dabbott (changes requested):

packages/app/src/containers/LayerStyleInspector.tsx
```diff
@@ -0,0 +1,302 @@
...
+import * as InspectorPrimitives from '../components/inspector/InspectorPrimitives';
+import useShallowArray from '../hooks/useShallowArray';
+import getMultiValue from '../utils/getMultiValue';
+import FillRow from '../components/inspector/FillRow';
+import BorderRow from '../components/inspector/BorderRow';
+import ShadowRow from '../components/inspector/ShadowRow';
+import withSeparatorElements from '../utils/withSeparatorElements';
+import NameInspector from './NameInspector';
+
+const StyleOpacityInspector = memo(function OpacityInspector() {
```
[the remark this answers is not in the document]
milg15: Oh! I was about to do that, join them but we need to change this lines in the inspector. 
image


By either sending them the selectors before or conditionally selecting. That's the only difference. the selector

packages/app/src/containers/ComponentsWindow.tsx
```diff
@@ -13,47 +14,18 @@ const Container = styled.main(({ theme }) => ({
...
-                  event.shiftKey ? 'intersection' : 'replace',
-                )
-              }
-            >
-              <ColorSwatch value={item.value} />
-            </GridView.Item>
-          );
-        })}
-      </GridView.Root>
+      {componentsTab === 'swatches' && <SwatchesGrid />}
```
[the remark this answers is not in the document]
milg15: image
I changed it like this

packages/app/src/containers/LayerStyleInspector.tsx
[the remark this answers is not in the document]
milg15: Something similiar to how NameInspector would work, where different types of objects can interact with it. So the selector shouldn't be inside the inpector I think

packages/noya-state/src/reducers/application.ts
```diff
@@ -1099,26 +782,92 @@ export function reducer(
...
+        const layerStyles = state.sketch.document.layerStyles ?? {
+          _class: 'sharedStyleContainer',
+          do_objectID: uuid(),
+          objects: [],
+        };
 
-      const ids = typeof id === 'string' ? [id] : id;
+        const style = {} as Sketch.Style;
+        const fill = {} as Sketch.Fill;
+        const contextSettings = {} as Sketch.GraphicsContextSettings;
```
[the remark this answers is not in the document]
milg15: I added a need style model instead. "style.json" 
image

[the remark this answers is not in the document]
milg15: image
Sorry just change it to this so it had an different id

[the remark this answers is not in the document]
milg15: I tried a way of changing the id of the style but it gives an error about being read only.

packages/app/src/containers/NameInspector.tsx
milg15: the issue I had with this was that I couldn't use the useCallback inside de `SwatchesInspector.tsx` and `LayerStylesInspector.tsx` hook because it gives this error 

> React Hook "useCallback" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?


Because I have a return null, for when the selectedSwatches or StyleLayer it's empty.

packages/noya-state/src/reducers/application.ts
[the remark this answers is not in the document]
dabbott: This looks good! I'm not sure about the readonly error, but maybe it will work if you use `produce`?

```ts
const sharedStyle: Sketch.SharedStyle = {
  _class: 'sharedStyle',
  do_objectID: uuid(),
  name: 'New Style Layer',
  value: produce(Models.style, style => {
    style.do_objectID = uuid()
  }),
};
```

packages/app/src/containers/NameInspector.tsx
dabbott: For cases like that, we can move the `useCallback` above the return, e.g.

```tsx
export default memo(function ComponentsInspectors() {
  /* ... */

  const handleNameChange = useCallback(
    (value: string) =>
      dispatch(
        'setSwatchName',
        selectedSwatches.map((v) => v.do_objectID),
        value,
      ),
    [dispatch, selectedSwatches],
  );

  if (selectedSwatches.length === 0) return null;

  return (
    <Fragment key="layout">
      <NameInspector
        names={selectedSwatches.map((v) => v.name)}
        onNameChange={handleNameChange}
      />
      <ColorPickerInspector />
      <Spacer.Vertical size={10} />
    </Fragment>
  );
});
```

packages/app/src/containers/LayerStyleInspector.tsx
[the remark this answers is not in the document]
dabbott: Ah, right. I can think of 2 options here:

1. We pass an array of `styles` or style ids to `OpacityInspector` as a prop, and move the selector outside the `OpacityInspector`
2. We use a different selector within `OpacityInspector` that returns an array of styles, and depending on the current tab it will either return layer styles or shared styles

I like option 1 longer term, but I think option 2 would be a much smaller change, and more consistent with the rest of this PR, since it will mirror how we currently handle actions. Something like...

```ts
// pseudocode, not sure if these functions all exist
export const getSelectedStyles = (state: ApplicationState): Sketch.Style[] => {
  const currentTab = getCurrentTab(state);
  return currentTab === 'canvas' 
    ? getSelectedLayers(state).flatMap(layer => layer.style ? [layer] : []) 
    : getSharedStyles(state).map(style => style.value)
};
```

What do you think?

packages/app/src/containers/NameInspector.tsx
milg15: ohh! I understand

packages/noya-state/src/reducers/application.ts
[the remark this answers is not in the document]
milg15: Oh! Now I know that ! 💯

packages/app/src/containers/LayerStyleInspector.tsx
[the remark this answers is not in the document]
milg15: I think option 2 it's better, I just added that one. We would have to change how NameInspector works to match it maybe but I just changed it. It looks cleaner and it follows a better pattern .

packages/noya-state/src/reducers/style.ts
```diff
@@ -0,0 +1,291 @@
+import Sketch from '@sketch-hq/sketch-file-format-ts';
+import produce from 'immer';
+import * as Models from '../models';
+
+export type SetNumberMode = 'replace' | 'adjust';
+
+export type StyleElementType = 'Fill' | 'Border' | 'Shadow';
+
+export type StyleShadowType = 'X' | 'Y' | 'Blur' | 'Spread';
```
dabbott, line 9: Maybe rename `StyleShadowType` => `StyleShadowProperty`, since I think "property" better describes x/y/blur/spread

packages/noya-state/src/selectors.ts
```diff
@@ -152,6 +174,16 @@ export const getSelectedColorSwatches = (
   );
 };
 
+export const getSelectedLayerStyle = (
```
dabbott, line 177: If this returns an array, should the name be plural? I.e. `getSelectedLayerStyles`

packages/noya-state/src/selectors.ts
```diff
@@ -103,6 +115,16 @@ export const getSelectedLayersExcludingDescendants = (
...
 };
 
+export const getSelectedStyles = (state: ApplicationState): Sketch.Style[] => {
+  const currentTab = getCurrentTab(state);
+
+  return currentTab === 'canvas'
+    ? getSelectedLayers(state).flatMap((layer) =>
+        layer.style ? [layer.style] : [],
+      )
+    : getSelectedLayerStyle(state).flatMap((style) => style.value);
```
dabbott, line 125: Can this be a regular `.map`? We only need `flatMap` to flatten arrays - in the first case I use it to remove undefined values, but I don't think we have that in this second case

packages/noya-state/src/reducers/application.ts
```diff
@@ -1099,26 +776,76 @@ export function reducer(
...
+          _class: 'sharedStyleContainer',
+          do_objectID: uuid(),
+          objects: [],
+        };
 
-      const ids = typeof id === 'string' ? [id] : id;
+        const sharedStyle: Sketch.SharedStyle = {
+          _class: 'sharedStyle',
+          do_objectID: uuid(),
+          name: 'New Style Layer',
```
dabbott, line 798: `'New Style Layer'` => `'New Layer Style'`

dabbott (changes requested):
This is looking great! Now it's just a few small things and then we're ready to merge 🚀

dabbott (approved):
Excellent! 🎉🎉🎉

PR #12, opened 2021-04-05
Refactor theme tab

dabbott:
The main change here is renaming the "Components" tab to the "Theme" tab. I borrowed the name "Components" from Sketch, but it's confusing, since we also have React components 😆

I also changed a few "container" components, which use `useApplicationState` or `useSelector`, to "presentational" components, which communicate everything via props. I want to convert a few more container components in the future, since presentational components are easier to reuse and easier to think about (smaller API surface area compared with `useApplicationState`).

In a future refactor, maybe we could make a utility function for sorting in `noya-utils`, since we do the same sort in a few places now.


packages/app/src/containers/SwatchesInspector.tsx
```diff
@@ -65,16 +37,37 @@ export default memo(function ComponentsInspectors() {
...
       <NameInspector
         names={selectedSwatches.map((v) => v.name)}
         onNameChange={handleNameChange}
       />
-      <ColorPickerInspector />
-      <Spacer.Vertical size={10} />
-    </Fragment>
+      <Divider />
+      <Container>
+        <ColorInspector
```
milg15, line 63: Oh I like it more like this, it also follows what I did before with the name! Make sense

PR #14, opened 2021-04-06
Layertheme inspector

milg15:
Add  a new theme image

Disabled if not selected
image

New actions! 
image

We should change the the in the prompt! But it works!

### DIFF
--- a/packages/app/src/containers/LayerThemeInspector.tsx
+++ b/packages/app/src/containers/LayerThemeInspector.tsx
@@ -10,15 +10,33 @@ import { Selectors } from 'noya-state';
 import { memo, useCallback, useMemo } from 'react';
 import { useTheme } from 'styled-components';
 import * as InspectorPrimitives from '../components/inspector/InspectorPrimitives';
-import { useSelector } from '../contexts/ApplicationStateContext';
+import {
+  useApplicationState,
+  useSelector,
+} from '../contexts/ApplicationStateContext';
 import useShallowArray from '../hooks/useShallowArray';
 
 const NO_LAYER_STYLE = 'none';
 
 export default memo(function LayerThemeInspector() {
+  const [, dispatch] = useApplicationState();
+
   const iconColor = useTheme().colors.icon;
 
   const sharedStyles = useShallowArray(useSelector(Selectors.getSharedStyles));
+  const selectedLayers = useShallowArray(
+    useSelector(Selectors.getSelectedLayers),
+  );
+
+  const selectedLayerStyleId = useMemo(
+    () =>
+      !selectedLayers.every(
+        (v) => v.sharedStyleID === selectedLayers[0].sharedStyleID,
+      )
+        ? undefined
+        : selectedLayers[0].sharedStyleID,
+    [selectedLayers],
+  );
 
   const layerStyleOptions = useMemo(
     () => [NO_LAYER_STYLE, ...sharedStyles.map((style) => style.do_objectID)],
@@ -33,6 +51,25 @@ export default memo(function LayerThemeInspector() {
     [sharedStyles],
   );
 
+  const onLayerStyleChange = useCallback(
+    (value) => {
+      const style = sharedStyles.find((s) => s.do_objectID === value);
+
+      if (style !== undefined) dispatch('setLayerStyle', value, style.value);
+      else dispatch('setLayerStyle', value, undefined);
+    },
+    [sharedStyles, dispatch],
+  );
+
+  const onAddLayerStyle = useCallback(() => {
+    const styleName = String(prompt('New Style Layout Name'));
+
+    dispatch('addLayerStyle', styleName, selectedLayers[0].style);
+  }, [selectedLayers, dispatch]);
+
+  /**
+   * When adding a new style  layout. Whay should happen when multiselecting?
+   */
   return (
     <>
       <InspectorPrimitives.Section>
@@ -43,10 +80,10 @@ export default memo(function LayerThemeInspector() {
         <InspectorPrimitives.Row>
           <Select
             id="theme-layer-style"
-            value={'No Layer Style'}
+            value={selectedLayerStyleId || 'No Layer Style'}
             options={layerStyleOptions}
             getTitle={getLayerStyleTitle}
-            onChange={() => {}}
+            onChange={onLayerStyleChange}
           />
         </InspectorPrimitives.Row>
         <Spacer.Vertical size={10} />
@@ -54,7 +91,7 @@ export default memo(function LayerThemeInspector() {
           <Button
             id="create-layer-style"
             tooltip="Create theme style from layer"
-            onClick={useCallback(() => {}, [])}
+            onClick={onAddLayerStyle}
           >
             <PlusIcon color={iconColor} />
           </Button>
--- a/packages/app/src/containers/LayerThemeInspector.tsx
+++ b/packages/app/src/containers/LayerThemeInspector.tsx
@@ -73,23 +73,18 @@ export default memo(function LayerThemeInspector() {
   }, [selectedLayers, dispatch]);
 
   const onRename = useCallback(() => {
-    const id = selectedLayers[0].sharedStyleID;
-
-    if (id !== undefined) {
-      const newName = prompt('Rename Layout style ');
+    if (selectedLayerStyleId !== undefined) {
+      const newName = prompt('Rename Layout style');
 
       if (newName !== null) {
-        dispatch('setLayerStyleName', id, newName);
+        dispatch('setLayerStyleName', selectedLayerStyleId, newName);
       }
     } else {
       alert('No Layer Selected');
     }
-  }, [selectedLayers, dispatch]);
+  }, [selectedLayerStyleId, dispatch]);
 
-  const onDetach = useCallback(
-    () => dispatch('setLayerStyle', undefined, undefined),
-    [dispatch],
-  );
+  const onDetach = useCallback(() => dispatch('setLayerStyle'), [dispatch]);
 
   const onUpdate = useCallback(
     () =>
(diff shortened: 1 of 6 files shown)


packages/noya-state/src/reducers/application.ts
```diff
@@ -791,16 +798,30 @@ export function reducer(
...
+              return style;
+            },
+          ),
         };
 
         layerStyles.objects.push(sharedStyle);
         state.sketch.document.layerStyles = layerStyles;
-        state.selectedLayerStyleIds = [sharedStyle.do_objectID];
+
+        if (style === undefined) {
```
dabbott, line 814: I find this `if / else` logic confusing, since it's not clear from this part of the code why we would do something fairly different if `style === undefined`.

I'd suggest:

```ts
if (currentTab === 'theme') {
  // Update selection
} else {
  // Modify layers
}
```

When I have more time, I want to revisit how we rely on `currentTab` for different behavior, since I think we can come up with something more clear

packages/noya-state/src/reducers/application.ts
```diff
@@ -791,16 +798,30 @@ export function reducer(
...
+          state.selectedLayerStyleIds = [sharedStyle.do_objectID];
+        } else {
+          accessPageLayers(state, pageIndex, layerIndexPaths).forEach(
+            (layer) => {
+              if (!layer.style) return;
```
dabbott, line 819: I don't think we need `if (!layer.style) return;`. If the style doesn't exist, we have a different problem 😂

packages/noya-state/src/reducers/application.ts
```diff
@@ -845,6 +866,46 @@ export function reducer(
...
     }
+    case 'setLayerStyle': {
+      const pageIndex = getCurrentPageIndex(state);
+      const layerIndexPaths = getSelectedLayerIndexPaths(state);
+
+      const [, id, style] = action;
+
+      return produce(state, (state) => {
+        accessPageLayers(state, pageIndex, layerIndexPaths).forEach((layer) => {
+          if (!layer.style) return;
```
dabbott, line 877: I don't think we need `if (!layer.style) return` here either

packages/noya-state/src/reducers/application.ts
```diff
@@ -845,6 +866,46 @@ export function reducer(
...
+
+          if (style) {
+            layer.sharedStyleID = id;
+            layer.style = produce(style, (style) => {
+              style.do_objectID = uuid();
+              return style;
+            });
+          } else {
+            layer.sharedStyleID = undefined;
```
dabbott, line 886: I'd typically go with `delete layer.sharedStyleID` rather than setting to undefined

packages/noya-state/src/reducers/application.ts
```diff
@@ -845,6 +866,46 @@ export function reducer(
...
+        });
+      });
+    }
+    case 'updateLayerStyle': {
+      const [, id, style] = action;
+
+      return produce(state, (state) => {
+        if (!style) return;
+
+        const sharedStyle = state.sketch.document.layerStyles.objects;
```
dabbott, line 897: This can be `sharedStyle` => `sharedStyles`

So below we can have `SStyle` => `sharedStyle`

packages/noya-designsystem/src/components/Button.tsx
```diff
@@ -59,11 +61,17 @@ function Button({
...
   onClick,
   children,
 }: ButtonRootProps) {
   const buttonElement = (
-    <ButtonElement id={id} active={active} onClick={onClick}>
+    <ButtonElement
+      id={id}
+      active={active}
+      disabled={disabled}
+      onClick={onClick}
```
dabbott, line 73: Can we prevent the `onClick` from firing in the case of a disabled button? (does that happen already?)

packages/app/src/containers/LayerThemeInspector.tsx
```diff
@@ -43,50 +118,54 @@ export default memo(function LayerThemeInspector() {
         <InspectorPrimitives.Row>
           <Select
             id="theme-layer-style"
-            value={'No Layer Style'}
+            value={selectedLayerStyleId || 'No Layer Style'}
```
dabbott, line 121: I know I wrote this, but can we use the constant here, `selectedLayerStyleId || NO_LAYER_STYLE`?

packages/app/src/containers/LayerThemeInspector.tsx
```diff
@@ -33,6 +51,63 @@ export default memo(function LayerThemeInspector() {
...
+    } else {
+      alert('No Layer Selected');
+    }
+  }, [selectedLayerStyleId, dispatch]);
+
+  const onDetach = useCallback(() => dispatch('setLayerStyle'), [dispatch]);
+
+  const onUpdate = useCallback(
+    () =>
+      selectedLayerStyleId
```
dabbott, line 91: We can use `if (!selectedLayerStyleId) return` here. We typically don't want a ternary `a ? b() : undefined` , where one branch has a side effect and the other doesn't

packages/app/src/containers/LayerThemeInspector.tsx
```diff
@@ -33,6 +51,63 @@ export default memo(function LayerThemeInspector() {
...
+        ? dispatch(
+            'updateLayerStyle',
```
dabbott, line 93: `updateLayerStyle` => `updateThemeStyle`?

packages/noya-state/src/reducers/application.ts
```diff
@@ -121,7 +121,8 @@ export type Action =
       mode?: SetNumberMode,
     ]
   | [type: 'addColorSwatch']
-  | [type: 'addLayerStyle']
+  | [type: 'addLayerStyle', name?: string, style?: Sketch.Style]
+  | [type: 'updateLayerStyle', id: string, style: Sketch.Style | undefined]
```
dabbott, line 125: For each of the new actions, can we clarify `id` by making it either `layerId` or `sharedStyleId`?

packages/app/src/containers/LayerThemeInspector.tsx
```diff
@@ -33,6 +51,63 @@ export default memo(function LayerThemeInspector() {
...
+      dispatch('addLayerStyle', styleName, selectedLayers[0].style);
+    }
+  }, [selectedLayers, dispatch]);
+
+  const onRename = useCallback(() => {
+    if (selectedLayerStyleId !== undefined) {
+      const newName = prompt('Rename Layout style');
+
+      if (newName !== null) {
+        dispatch('setLayerStyleName', selectedLayerStyleId, newName);
```
dabbott, line 80: `setLayerStyleName` => `setThemeStyleName`?

dabbott (changes requested):
This is looking good! My main feedback is around naming. I think we should make it more clear whether actions modify layers vs. theme styles. So I proposed a bunch of new action names, and renaming `id` to e.g. `layerId` or `sharedStyleId` wherever possible.

I generally like the name "theme style" better than "shared style", however since `sharedStyleId` is a Sketch property, we shouldn't change that. So we'll probably have a mix of both, but I think that's OK for now. And since I find "theme layer style" confusing, I think any layer styles in the theme can just be called "theme styles".

I'll probably have some more feedback once we make those changes and it's easier to compare the actions all together

packages/app/src/containers/LayerThemeInspector.tsx
```diff
@@ -33,6 +51,63 @@ export default memo(function LayerThemeInspector() {
...
+        dispatch('setLayerStyle', value, undefined);
+      }
+    },
+    [sharedStyles, dispatch],
+  );
+
+  const onAdd = useCallback(() => {
+    const styleName = prompt('New Style Layout Name');
+
+    if (styleName != null) {
```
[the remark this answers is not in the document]
milg15: I have a case in the action where if the string is '' it will call it New Style Layer

packages/noya-designsystem/src/components/Button.tsx
milg15: it happens already, if you see in the html, the button it's disabled. I think the button styled components is already handling it.  
image

packages/app/src/containers/LayerThemeInspector.tsx
```diff
@@ -33,6 +51,48 @@ export default memo(function LayerThemeInspector() {
...
+    const newName = prompt('Rename Layout style');
+
+    if (!newName) return;
+    dispatch('setThemeStyleName', selectedLayerStyleId, newName);
+  }, [selectedLayerStyleId, dispatch]);
+
+  const onDetach = useCallback(() => dispatch('setThemeStyle'), [dispatch]);
+
+  const onUpdate = useCallback(() => {
+    console.log('why?');
```
dabbott, line 80: The console.log 😂

packages/noya-state/src/reducers/application.ts
```diff
@@ -845,6 +870,49 @@ export function reducer(
...
+      const layerIndexPaths = getSelectedLayerIndexPaths(state);
+
+      const [, id] = action;
+
+      return produce(state, (state) => {
+        accessPageLayers(state, pageIndex, layerIndexPaths).forEach((layer) => {
+          const style =
+            state.sketch.document.layerStyles?.objects.find(
+              (s) => s.do_objectID === id,
+            ) || null;
```
dabbott, line 884: I don't think we'd need an `|| null` here

dabbott (changes requested):
This is great! I think renaming helped a lot, and I think the actions you have now are good. Just a couple small things left.

There are a some other potential improvements that come to mind, but we don't need to try these now, cause I'm not sure about them yet:

- Potentially refactor theme styles stuff into a separate reducer
- Potentially add more actions rather than overloading with optional parameters, e.g. a `"detachThemeStyle"` rather than passing undefined to `"setThemeStyle"`

dabbott (approved):
🚀

PR #16, opened 2021-04-07
implement history stack

souporserious:
Implements a history stack so actions on the file can be applied from the past or future if available. Right now the timeout heuristic does not take into account dragging, so multiple actions can be recorded during a drag event like selecting a color or moving a layer.

### DIFF
--- a/packages/app/src/contexts/ApplicationStateContext.tsx
+++ b/packages/app/src/contexts/ApplicationStateContext.tsx
@@ -33,6 +33,7 @@ export const useRawApplicationState = (): ApplicationStateContextValue => {
 type WrappedApplicationState = ApplicationState & {
   undoDisabled: boolean;
   redoDisabled: boolean;
+  fullHistory: HistoryState;
 };
 
 /**
@@ -71,6 +72,7 @@ export const useApplicationState = (): [
     return [
       {
         ...state.present,
+        fullHistory: state,
         undoDisabled: state.past.length === 0,
         redoDisabled: state.future.length === 0,
       },
--- a/packages/app/src/contexts/ApplicationStateContext.tsx
+++ b/packages/app/src/contexts/ApplicationStateContext.tsx
@@ -1,9 +1,4 @@
-import {
-  Action,
-  ApplicationState,
-  HistoryAction,
-  HistoryState,
-} from 'noya-state';
+import { ApplicationState, HistoryAction, HistoryState } from 'noya-state';
 import { createContext, useCallback, useContext, useMemo } from 'react';
 import { useGlobalInputBlurTrigger } from 'noya-designsystem';
 
@@ -18,7 +13,7 @@ const ApplicationStateContext = createContext<
 
 export const ApplicationStateProvider = ApplicationStateContext.Provider;
 
-type Dispatcher = (...args: Action) => void;
+type Dispatcher = (...args: HistoryAction) => void;
 
 /**
  * This should only be used to propagate state between React reconcilers
@@ -41,7 +36,10 @@ export const useRawApplicationState = (): ApplicationStateContextValue => {
  * Only "container" components should use this, while "presentational" components
  * should instead be passed their data via props.
  */
-export const useApplicationState = (): [ApplicationState, Dispatcher] => {
+export const useApplicationState = (): [
+  ApplicationState & { undoDisabled: boolean; redoDisabled: boolean },
+  Dispatcher,
+] => {
   const value = useRawApplicationState();
   const trigger = useGlobalInputBlurTrigger();
 
@@ -49,7 +47,7 @@ export const useApplicationState = (): [ApplicationState, Dispatcher] => {
 
   // Simplify the dispatch function by flattening our Action tuple
   const wrappedDispatch: Dispatcher = useCallback(
-    (...args: Action) => {
+    (...args: HistoryAction) => {
       // When changing selection, trigger any pending updates in input fields
       if (
         args[0] === 'selectLayer' ||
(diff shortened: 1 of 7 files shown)


packages/noya-state/src/reducers/history.ts
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export type HistoryAction = [type: 'undo'] | [type: 'redo'] | Action;
+
+const FILE_CHANGED_TIMEOUT = 300;
+
+export function historyReducer(state: HistoryState, action: HistoryAction) {
+  const currentPresent = state.present;
+  const [actionType] = action;
+  switch (actionType) {
```
dabbott, line 24: We should do `switch (action[0]) { ... }` here. Defining `actionType` as a separate var will prevent typescript from narrowing the type of `action`

packages/noya-state/src/reducers/history.ts
```diff
@@ -0,0 +1,99 @@
...
+    case 'undo':
+      return produce(state, (state) => {
+        if (state.past.length > 0) {
+          const newPresent = state.past.pop() as HistoryEntry;
```
dabbott, line 28: We should figure some way of avoiding the `as` cast here, even if it means doing some work before we call `produce`

packages/noya-state/src/reducers/history.ts
```diff
@@ -0,0 +1,99 @@
+import produce from 'immer';
+import { ApplicationState, Action, reducer } from './application';
+import { createInitialInteractionState } from './interaction';
+
+export type HistoryEntry = {
+  actionType: Action[0];
+  timestamp: number;
+  state: ApplicationState;
```
dabbott, line 8: Can `state` just be a `SketchFile`? Is there anything else we actually want to preserve when we undo/redo (for now)?

packages/noya-state/src/reducers/history.ts
dabbott: Probably we want to do `if (state.past.length === 0) { return state }` before `produce` regardless

packages/noya-state/src/reducers/history.ts
```diff
@@ -0,0 +1,99 @@
...
+              createHistoryEntry(newPresent.actionType, currentPresent),
+            );
+            state.present = newPresent.state;
+          }
+        }
+      });
+    default:
+      const newPresent = reducer(currentPresent, action as Action);
+      return produce(state, (state) => {
+        const sketchFileChanged =
```
dabbott, line 52: This is going to be a really expensive comparison. Can we look into why `!==` isn't working? Maybe we can come up with something better

packages/noya-state/src/reducers/history.ts
```diff
@@ -0,0 +1,99 @@
...
+          if (newPresent) {
+            state.past.push(
+              createHistoryEntry(newPresent.actionType, currentPresent),
+            );
+            state.present = newPresent.state;
+          }
+        }
+      });
+    default:
+      const newPresent = reducer(currentPresent, action as Action);
```
dabbott, line 50: I think when we update the switch, we won't need `as` here

packages/noya-state/src/reducers/history.ts
```diff
@@ -0,0 +1,99 @@
...
+        const historyEntry = createHistoryEntry(actionType, {
+          ...currentPresent,
+          interactionState: createInitialInteractionState(),
+        });
+
+        const pushHistoryEntry = () => {
+          state.past.push(historyEntry);
+        };
+
+        if (sketchFileChanged) {
```
dabbott, line 65: To improve the double `else` (I didn't test this, but it typechecks OK at least, after fixing the switch):

```ts
const mergableEntry = getMergableHistoryEntry(state, action[0]);

return produce(state, (state) => {
  // ...

  if (sketchFileChanged) {
    if (mergableEntry) {
      state.past[state.past.length - 1] = {
        ...historyEntry,
        state: mergableEntry.state,
      };
    } else {
      state.past.push(historyEntry);
    }

    state.future = [];
  }

  // ...
});
```

```ts
function getMergableHistoryEntry(
  state: HistoryState,
  actionType: Action[0],
): HistoryEntry | undefined {
  if (state.past.length > 0) return;

  const newTimestamp = Date.now();
  const previousEntry = state.past[state.past.length - 1];

  if (
    actionType !== previousEntry.actionType ||
    newTimestamp - previousEntry.timestamp > FILE_CHANGED_TIMEOUT
  ) {
    return;
  }

  return previousEntry;
}
```

dabbott (changes requested):
This is looking good! Since this is very core functionality, I want to iterate on it a bit before we merge. 

Can you rebase/merge with main? There's a new eslint rule I want to pick up.

packages/app/src/contexts/ApplicationStateContext.tsx
```diff
@@ -30,21 +30,29 @@ export const useRawApplicationState = (): ApplicationStateContextValue => {
   return value;
 };
 
+type WrappedApplicationState = ApplicationState & {
```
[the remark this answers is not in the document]
souporserious: Love this, I'll get this updated!

packages/noya-state/src/reducers/history.ts
souporserious: Ah, I was wondering why it was complaining 😅

packages/noya-state/src/reducers/history.ts
souporserious: I think so, should we store selection state similar to other design tools?

packages/noya-state/src/reducers/history.ts
souporserious: Yeah, I was weary of using stringify since it's slow, I'll dig deeper on why we can't compare.

packages/noya-state/src/reducers/history.ts
souporserious: Leaving this for now per our Slack convo.

dabbott (approved):
Looks great! It's very cool to see it all come together 🔥

Issue #15, opened 2021-04-07
Add "delete layer" button


dabbott:
Add a button to the toolbar for deleting any selected layer(s)

Details:
- There should be a corresponding `deleteLayer` action, that takes a layer id, or an array of layer ids, and deletes them from the layer hierarchy
- This can use the `TrashIcon` component for the icon

milg15:
Could it also be when pressing the del key?

dabbott:
We definitely need keyboard shortcuts everywhere, but it's going to be a big task, so I've been putting it off 😂

milg15:
This branch was created from the other branch I was working on 😅  So it has more changes than it should. I'll be moving the branch to be in sync with main instead. I'll just delete the other one and create a new one.

PR #18, opened 2021-04-08
Delete layer button 🗑️ 

milg15:
okey new name but same branch 👍 solves #15

### DIFF
--- a/packages/app/src/containers/SwatchesToolbar.tsx
+++ b/packages/app/src/containers/SwatchesToolbar.tsx
@@ -57,7 +57,7 @@ export default function SwatchesToolbar() {
     dispatch,
   ]);
 
-  const addThemeStyle = useCallback(() => dispatch('addThemeStyle'), [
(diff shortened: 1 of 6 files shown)


packages/app/src/containers/SwatchesToolbar.tsx
```diff
@@ -94,10 +103,33 @@ export default function SwatchesToolbar() {
           >
             <PlusIcon />
           </Button>
+          <Spacer.Horizontal size={4} />
+          <Button
+            id="add-style"
+            onClick={() => {
```
dabbott, line 109: We should add a `useCallback` here. This is the function that would need `useCallback` most, since without it, the `Button` will always re-render

dabbott (changes requested):

milg15:
Okey I had to do some sync, it was working weird 😆 I hope it keeps working correctly.

dabbott:
After I merged your other branch, it looks like there are a few conflicts. Can you try resolving those?

milg15:
#20 Third time's a Charm

Issue #24, opened 2021-04-14
Add context menu to grid items


dabbott:
## Feature

Let's add a right-click context menu to items in the grid on the theme tab.

![Screen Shot 2021-04-14 at 3 53 57 PM](https://user-images.githubusercontent.com/1198882/114791071-abbebc00-9d3a-11eb-94a3-0ad0579ebbce.png)

To keep this issue small, let's just implement `Delete` in the menu, for now. We can add the other menu items in another issue.

## Implementation

We should use `ContextMenu` from the `noya-designsystem` package. It can be rendered similar to:

```tsx
<ContextMenu.Root<MenuItemType>
  items={menuItems}
  onSelect={onSelectMenuItem}
>
  {/* render a the content of the grid item here */}
</ContextMenu.Root>
```

**Important**: `ContextMenu` should go in `GridView.tsx`, **not** in SwatchesGrid, LayerStylesGrid, etc

We will need to wrap `GridItem` in `forwardRef` to get this to work. You can hopefully refer to `ListView.Row` for a lot of this (and it should be a bit simpler, since we don't need to support sorting).

Each `GridItem` should take these 3 props, just like `ListView.Row`:

https://github.com/noya-app/noya/blob/7ec3094af2092516600c2eab7a660a0dafd3048a/packages/noya-designsystem/src/components/ListView.tsx#L153-L155

milg15:
I think I could implement more actions in this issue.

I am not really sure where to use the `ref` it gives me an error when I add it. The action works correcty I believe (The copy was me trying to see if I could add more)
![image](https://user-images.githubusercontent.com/16924758/114797116-12060780-9d58-11eb-808e-c2ae43c4c107.png)

dabbott:
If you want you could add `duplicate` and `rename` in this issue. After we get this one in, I'm going to make a new one about grouping, so we'll come back to that next. Generally I prefer smaller PRs though, cause they're more focused, and they're also less likely to cause merge conflicts.

The ref should probably be passed to `SwatchContainer` (in the current code), but if it's working I can take a look at what you have first

PR #25, opened 2021-04-15
delete from context menu

milg15:
solves #24

### DIFF
--- a/packages/noya-designsystem/src/components/GridView.tsx
+++ b/packages/noya-designsystem/src/components/GridView.tsx
@@ -1,6 +1,13 @@
-import { memo, useCallback, ReactNode } from 'react';
+import {
+  memo,
+  ForwardedRef,
+  forwardRef,
+  ReactNode,
+  Ref,
+  useCallback,
+} from 'react';
 import styled from 'styled-components';
-import { Spacer } from '..';
+import { Spacer, ContextMenu } from '..';
 
 const Grid = styled.div(({ theme }) => ({
   flex: 1,
@@ -40,7 +47,7 @@ const ItemContainer = styled.div<{ selected: boolean }>(
   }),
 );
 
-const SwatchContainer = styled.div(({ theme }) => ({
+const GridContainer = styled.div(({ theme }) => ({
   display: 'flex',
   flexDirection: 'column',
 }));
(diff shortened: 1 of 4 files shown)


packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -64,37 +71,76 @@ const ItemDescription = styled.span(({ theme }) => ({
...
+      (event: React.MouseEvent) => {
+        event.stopPropagation();
+        event.preventDefault();
+
+        onClick?.(event);
+      },
+      [onClick],
+    );
+
+    const renderContent = (ref: Ref<HTMLLIElement>) => {
```
dabbott, line 111: We used the `renderContent` function to help handle sorting in the ListView, but since we don't do sorting here, we don't need to wrap everything in a function

milg15:
Where it's the ref necessary ?

packages/noya-designsystem/src/components/GridView.tsx
milg15: Solved it, I had the wrong type for the ref

dabbott:
Since `GridContainer` is a div, it'd need to be `ForwardedRef<HTMLDivElement>`, but if that's not the issue then I'm not sure

milg15:
Maybe we can close this PR as it is if there isn't any errors, so we keep it small as you say and work the other actions on a new issue. Since it would require new actions and I prefer to not touch the `application.ts` file here since it always gives me errors when merging 😆

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -64,37 +64,74 @@ const ItemDescription = styled.span(({ theme }) => ({
   textOverflow: 'ellipsis',
 }));
 
+type MenuItemType = 'delete' | 'copy';
+
 interface ItemProps {
-  children?: ReactNode;
-  title: string;
+  id?: string;
```
dabbott, line 70: Ideally, the `MenuItemType` should be generic here. If `ItemProps` and `GridItem` are generic (`<MenuItemType extends string>`) then they can work with other kinds of menu items. We use generics in `ListView.Row`:

```ts
export interface ListViewRowProps<MenuItemType extends string = string> { ... }

function ListViewRow<MenuItemType extends string>(...) { ... }
```

In this case it will be:

```ts
export interface ItemProps<MenuItemType extends string = string> { ... }

function GridItem<MenuItemType extends string>(...) { ... }
```

You may also have to render it as `<GridItem<MenuItemType> ... />` in the other grids

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -64,37 +64,74 @@ const ItemDescription = styled.span(({ theme }) => ({
   textOverflow: 'ellipsis',
 }));
 
+type MenuItemType = 'delete' | 'copy';
```
dabbott, line 67: (See other comment) So we shouldn't have any specific item values like "delete" or "copy" in this file

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -64,37 +64,74 @@ const ItemDescription = styled.span(({ theme }) => ({
...
+
 interface ItemProps {
-  children?: ReactNode;
-  title: string;
+  id?: string;
+  title?: string;
```
dabbott, line 71: Does `title` need to be optional?

dabbott (changes requested):
Great! This is what I had in mind for this PR

packages/noya-designsystem/src/components/GridView.tsx
milg15: I think not, I changed it for other reason. I'll keep it as it was now.

dabbott (approved):

PR #26, opened 2021-04-15
Update style from inspector and consistent file names

### LINKED ISSUE #19: Updating a theme style should update all layers using that style
## Feature

When the user changes a theme style, we want to update all layers using that style to have the new properties.

For example, suppose the user changes the fill color of this "Success" style:

Screen Shot 2021-04-11 at 12 57 28 PM

We should find any layer using the "Success" style and replace the layer's entire style with the updated "Success" style.

Similarly, if the user deletes the "Success" style, we should detach that style from any layers using it.

## Implementation

We should do this as part of any action where we support modifying theme styles. I think right now that's `"updateThemeStyle"` and `"setThemeStyle"` (to handle deletions).

I'm thinking the algorithm to update layers looks like this:

- For each page in our document (`state.sketch.pages`):
  - Use `Layers.findAllIndexPaths` to find all layers using the `sharedStyleId` of e.g. the "Success" style
  - Iterate over each layer using `accessPageLayers`
  - Replace the existing `layer.style` with the `sharedStyleId` (assigning a new `uuid()` like we do elsewhere)

And when a theme style is deleted:

- For each page in our document (`state.sketch.pages`):
  - Use `Layers.findAllIndexPaths` to find all layers using the `sharedStyleId` of e.g. the "Success" style
  - Iterate over each layer using `accessPageLayers`
  - Delete the `sharedStyleId`

milg15:
fix #19 Now it updates all layers when you change the theme style from the inspector. 

I also changed the name of SwatchesToolbar to ThemeToolbar so all files have related names (ThemeWindows, ThemeInspector)


dabbott (approved):
Looks good! Maybe at some point we'll make a separate function for this, since the logic is very similar to that other action (I forget which off the top of my head)

PR #29, opened 2021-04-15
group and ungroup actions, change name

milg15:
Handles groups and ungroups but stills need improvements when it comes to nested groups.

### DIFF
--- a/packages/app/src/components/inspector/NameInspector.tsx
+++ b/packages/app/src/components/inspector/NameInspector.tsx
@@ -1,4 +1,5 @@
 import { InputField, Spacer } from 'noya-designsystem';
+import { delimitedPath } from 'noya-utils';
 import { memo } from 'react';
 import * as InspectorPrimitives from './InspectorPrimitives';
 
@@ -8,14 +9,11 @@ interface Props {
 }
 
 export default memo(function NameInspector({ names, onNameChange }: Props) {
-  const firstName = names[0].split('/').pop() || '';
+  const firstName = delimitedPath.basename(names[0]);
 
   const name =
     names.length > 1 &&
-    !names.every((v: string) => {
-      const n = v.split('/');
-      return n[n.length - 1] === firstName;
-    })
+    !names.every((name: string) => delimitedPath.basename(name) === firstName)
       ? undefined
       : firstName;
 
--- a/packages/app/src/components/inspector/NameInspector.tsx
+++ b/packages/app/src/components/inspector/NameInspector.tsx
@@ -1,5 +1,4 @@
 import { InputField, Spacer } from 'noya-designsystem';
-import { delimitedPath } from 'noya-utils';
 import { memo } from 'react';
 import * as InspectorPrimitives from './InspectorPrimitives';
 
@@ -9,11 +8,10 @@ interface Props {
 }
 
 export default memo(function NameInspector({ names, onNameChange }: Props) {
-  const firstName = delimitedPath.basename(names[0]);
+  const firstName = names[0];
 
   const name =
-    names.length > 1 &&
-    !names.every((name: string) => delimitedPath.basename(name) === firstName)
+    names.length > 1 && !names.every((name: string) => name === firstName)
       ? undefined
       : firstName;
 
--- a/packages/app/src/components/inspector/NameInspector.tsx
+++ b/packages/app/src/components/inspector/NameInspector.tsx
@@ -8,7 +8,7 @@ interface Props {
 }
 
 export default memo(function NameInspector({ names, onNameChange }: Props) {
-  const firstName = names[0];
+  const firstName = names[0].split('/').pop() || '';
 
   const name =
     names.length > 1 && !names.every((v: string) => v === firstName)
--- a/packages/app/src/components/inspector/NameInspector.tsx
+++ b/packages/app/src/components/inspector/NameInspector.tsx
@@ -11,7 +11,11 @@ export default memo(function NameInspector({ names, onNameChange }: Props) {
   const firstName = names[0].split('/').pop() || '';
 
   const name =
-    names.length > 1 && !names.every((v: string) => v === firstName)
+    names.length > 1 &&
+    !names.every((v: string) => {
+      const n = v.split('/');
+      return n[n.length - 1] === firstName;
+    })
       ? undefined
       : firstName;
 
--- a/packages/app/src/components/theme/SwatchesGrid.tsx
+++ b/packages/app/src/components/theme/SwatchesGrid.tsx
@@ -18,13 +18,13 @@ function flatten(
 }[] {
   const path = (parent ? parent + '/' : '') + SwatchGroup.name;
 
-  return Array.prototype.concat.apply(
+  return [
     {
       path: path,
       swatches: SwatchGroup.swatches,
     },
-    SwatchGroup.children.map((x) => flatten(x, path)),
-  );
+    ...SwatchGroup.children.flatMap((x) => flatten(x, path)),
+  ];
 }
 
 interface Props {
--- a/packages/app/src/components/theme/SwatchesGrid.tsx
+++ b/packages/app/src/components/theme/SwatchesGrid.tsx
@@ -29,29 +29,31 @@ export default memo(function SwatchesGrid({
 
   return (
     <GridView.Root onClick={() => onSelectSwatch(undefined, 'replace')}>
-      {sortedSwatches.map((item) => {
-        const color = sketchColorToRgba(item.value);
-        const hex = rgbaToHex(color);
-        const alphaPercent = `${Math.round(color.a * 100)}%`;
+      <GridView.Section>
+        {sortedSwatches.map((item) => {
+          const color = sketchColorToRgba(item.value);
+          const hex = rgbaToHex(color);
+          const alphaPercent = `${Math.round(color.a * 100)}%`;
 
-        return (
-          <GridView.Item
-            id={item.do_objectID}
-            key={item.do_objectID}
-            title={item.name}
-            subtitle={`${hex} — ${alphaPercent}`}
-            selected={selectedSwatchIds.includes(item.do_objectID)}
-            onClick={(event: React.MouseEvent) =>
-              onSelectSwatch(
-                item.do_objectID,
-                event.shiftKey ? 'intersection' : 'replace',
-              )
-            }
-          >
-            <ColorSwatch value={item.value} />
-          </GridView.Item>
-        );
-      })}
+          return (
+            <GridView.Item
+              id={item.do_objectID}
+              key={item.do_objectID}
+              title={item.name}
+              subtitle={`${hex} — ${alphaPercent}`}
+              selected={selectedSwatchIds.includes(item.do_objectID)}
+              onClick={(event: React.MouseEvent) =>
+                onSelectSwatch(
+                  item.do_objectID,
+                  event.shiftKey ? 'intersection' : 'replace',
+                )
+              }
+            >
+              <ColorSwatch value={item.value} />
+            </GridView.Item>
+          );
+        })}
+      </GridView.Section>
     </GridView.Root>
   );
 });
--- a/packages/app/src/components/theme/SwatchesGrid.tsx
+++ b/packages/app/src/components/theme/SwatchesGrid.tsx
@@ -11,49 +11,74 @@ interface Props {
   onSelectSwatch: (id?: string, selectionType?: SelectionType) => void;
 }
 
+const sortArray = (array: any[], key: string) =>
+  [...array].sort((a, b) => {
+    const aName = a[key].toUpperCase();
+    const bName = b[key].toUpperCase();
+
+    return aName > bName ? 1 : aName < bName ? -1 : 0;
+  });
+
 export default memo(function SwatchesGrid({
   swatches,
   selectedSwatchIds,
   onSelectSwatch,
 }: Props) {
-  const sortedSwatches = useMemo(
-    () =>
-      [...swatches].sort((a, b) => {
-        const aName = a.name.toUpperCase();
-        const bName = b.name.toUpperCase();
+  const swatchesGrouped = useMemo(() => {
+    const group: { title: string; items: Sketch.Swatch[] }[] = [];
+    const indexes: { [key: string]: number } = {};
 
-        return aName > bName ? 1 : aName < bName ? -1 : 0;
-      }),
-    [swatches],
-  );
+    swatches.forEach((swatch) => {
+      const groupTitle = swatch.name.split('/').slice(0, -1).join('/');
+
+      if (!(groupTitle in indexes)) {
+        indexes[groupTitle] = group.length;
+        group.push({ title: groupTitle, items: [] });
+      }
+
+      group[indexes[groupTitle]].items.push(swatch);
+    });
+
+    return sortArray(group, 'title');
+  }, [swatches]);
 
   return (
     <GridView.Root onClick={() => onSelectSwatch(undefined, 'replace')}>
-      <GridView.Section>
-        {sortedSwatches.map((item) => {
-          const color = sketchColorToRgba(item.value);
-          const hex = rgbaToHex(color);
-          const alphaPercent = `${Math.round(color.a * 100)}%`;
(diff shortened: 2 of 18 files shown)


milg15:
I think there is some code that could be move to util. The sorting array one and the flattening array might be good to move to a different file maybe as a selector(?).

packages/app/src/containers/SwatchesGroups.tsx
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export default memo(function SwatchesGroups() {
+  const [state, dispatch] = useApplicationState();
+
+  const swatches = useShallowArray(useSelector(Selectors.getSharedSwatches));
+  const selectedGroup = state.selectedGroupSwatch;
+
+  const flatSwatchGroup = useMemo(() => {
+    const flat = flatten(createSwatchTree(swatches), '', -1);
```
dabbott, line 103: What does the -1 do here?

packages/app/src/containers/SwatchesGroups.tsx
```diff
@@ -0,0 +1,146 @@
...
+  });
+  return root;
+}
+
+function flatten(
+  swatch: SwatchGroup,
+  parent: string,
+  depth: number,
+): SwatchTitles[] {
+  const full = (parent ? parent + '/' : '') + swatch.name;
```
dabbott, line 84: Maybe rename `full` => `path`

packages/noya-state/src/reducers/application.ts
```diff
@@ -1000,9 +1007,11 @@ export function reducer(
             : state.sketch.document.layerStyles?.objects ?? [];
 
         array.forEach((object: Sketch.Swatch | Sketch.SharedStyle) => {
-          if (ids.includes(object.do_objectID)) {
-            object.name = name;
-          }
+          if (!ids.includes(object.do_objectID)) return;
+          const group = object.name.split('/').slice(0, -1).join('/');
```
dabbott, line 1011: Can this be simplified using the new `dirname` util?

packages/noya-state/src/reducers/application.ts
```diff
@@ -1114,6 +1123,33 @@ export function reducer(
...
+      const [, id, value] = action;
+
+      const ids = typeof id === 'string' ? [id] : id;
+
+      return produce(state, (state) => {
+        const array = state.sketch.document.sharedSwatches?.objects ?? [];
+
+        array.forEach((object: Sketch.Swatch) => {
+          if (!ids.includes(object.do_objectID)) return;
+          const values = object.name.split('/');
```
dabbott, line 1142: Can this function be simplified with the new `basename` and `dirname` utils?

packages/app/src/containers/SwatchesGroups.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const aName = a[key].toUpperCase();
+    const bName = b[key].toUpperCase();
+
+    return aName > bName ? 1 : aName < bName ? -1 : 0;
+  });
+
+type SwatchGroup = {
+  name: string;
+  swatches: Sketch.Swatch[];
+  children?: SwatchGroup[];
```
dabbott, line 34: Does it help to make children optional?

dabbott (changes requested):
I like the idea of making some utils. Let's make:

- `noya-utils/src/sortBy` (rename `sortArray` => `sortBy`) - Anything in `noya-utils` needs tests! You can run tests just for this package with `yarn test noya-utils`
- `app/src/utils/createSwatchTree` - We will make this generic soon but for now we can just move it. I don't think it's general enough to move to the `noya-utils`
- I'm not sure about flatten yet, but you could move it if you'd like (into app/src/utils/). And maybe rename to `flattenTree`.

I added 2 new utils directly to this branch: `dirname` and `basename`. These mirror the node.js APIs with the same names. We can hopefully use these to simplify (some of) the `.split('/')` and `.join('/')` stuff. You can take a look at the tests to see how they work. I used `basename` in `NameInspector` to test it too.

I also added a pre-commit hook to prevent pushing to GitHub unless TypeScript compiles successfully.

I might have another round of feedback once we make these updates.

**Remember to pull, since I made changes on this branch**

packages/app/src/containers/SwatchesGroups.tsx
milg15: I was optional for how a function was working before but now it doesn't need to be👍

packages/noya-designsystem/src/components/GridView.tsx
```diff
@@ -147,12 +170,40 @@ function GridViewRoot({ children, onClick }: GridViewRootProps) {
...
+interface GridViewSectionHeaderProps {
+  title: string;
+}
+
+function GridViewSectionHeader({ title }: GridViewSectionHeaderProps) {
+  const groupedTitle = title.split('/');
+
+  return (
+    <SectionHeaderContainer>
+      {groupedTitle.map((title, index) => {
```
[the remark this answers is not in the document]
milg15: Oh! Awesome! I added the function but had to run it with this path

`import withSeparatorElements from '../../../app/src/utils/withSeparatorElements';`
😅 

I don't think it's optimal, it's there any better way of reaching this file?

packages/app/src/components/theme/SwatchesGrid.tsx
```diff
@@ -1,55 +1,131 @@
...
-  const sortedSwatches = useMemo(
-    () =>
-      [...swatches].sort((a, b) => {
-        const aName = a.name.toUpperCase();
-        const bName = b.name.toUpperCase();
-
-        return aName > bName ? 1 : aName < bName ? -1 : 0;
-      }),
-    [swatches],
+  const swatchesGrouped = useMemo(() => {
```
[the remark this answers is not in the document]
milg15: Yes! I'll join with the new swatch group file

packages/noya-state/src/reducers/application.ts
milg15: yes !

milg15:
I still need to add more test for the sortBy. I think it's better now that the createSwatchTree it's now separate. 

I kept the flatten functions in the files because they return differen results in my opinion.

packages/noya-designsystem/src/components/GridView.tsx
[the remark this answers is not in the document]
dabbott: Good catch! I forgot that was in another package. Let's move withSeparatorElements.ts => `noya-designsystem/src/utils/withSeparatorElements.ts` and export it from `noya-designsystem/src/index.tsx`

packages/app/src/components/inspector/NameInspector.tsx
```diff
@@ -8,10 +9,11 @@ interface Props {
 }
 
 export default memo(function NameInspector({ names, onNameChange }: Props) {
-  const firstName = names[0];
+  const firstName = delimitedPath.basename(names[0]);
```
dabbott, line 12: I just realized we're handling `names` and `onNameChange` inconsistently:

- `names` are all full paths, like `["group/nestedGroup/a"]`
- `onNameChange` only gets called with the base name, `"a"`.

I think we should use either full path or base name for both props. `NameInspector` doesn't need to know the full path, cause it doesn't show it, so I would transform `names` into an array of base name, `["a"]`, before passing it in as a prop. It might mean a little duplicated code in the other inspectors, but we can improve that in other ways

packages/app/src/components/theme/SwatchesGrid.tsx
```diff
@@ -1,55 +1,136 @@
...
+      }
+    },
+    [onSelectSwatch, onGroupSwatch, onDeleteSwatch, selectedSwatchIds],
+  );
 
-        return aName > bName ? 1 : aName < bName ? -1 : 0;
-      }),
+  const handleOnContextMenu = useCallback(
+    (id: string) => {
+      if (selectedSwatchIds.length && selectedSwatchIds.includes(id)) return;
```
dabbott, line 79: Do we need to check `selectedSwatchIds.length` here?

packages/app/src/components/theme/SwatchesGrid.tsx
```diff
@@ -1,55 +1,136 @@
...
+function flatten(
+  swatch: SwatchGroup,
+  parent: string,
+): {
+  path: string;
+  swatches: Sketch.Swatch[];
+}[] {
+  const path = (parent ? parent + '/' : '') + swatch.name;
+
+  return Array.prototype.concat.apply(
```
dabbott, line 20: Can we do

```ts
return [
  {
    path: path,
    swatches: swatch.swatches,
  },
  ...swatch.children.map((x) => flatten(x, path)),
);
```

packages/app/src/components/theme/SwatchesGrid.tsx
```diff
@@ -1,55 +1,136 @@
...
+import { ContextMenu, GridView, sketchColorToRgba } from 'noya-designsystem';
 import { SelectionType } from 'noya-state';
-import { memo, useMemo } from 'react';
+import { memo, useMemo, useCallback, Fragment } from 'react';
+import { SwatchGroup, createSwatchTree } from '../../utils/createSwatchTree';
+import { sortBy } from 'noya-utils';
 import ColorSwatch from './ColorSwatch';
 
+function flatten(
+  swatch: SwatchGroup,
```
dabbott, line 12: Rename `swatch` => `swatchGroup`

packages/app/src/components/theme/SwatchesGrid.tsx
```diff
@@ -1,55 +1,136 @@
...
-            <ColorSwatch value={item.value} />
-          </GridView.Item>
+          <Fragment key={index}>
+            {group.path && <GridView.SectionHeader title={group.path} />}
+            <GridView.Section>
+              {sortedSwatches.map((item) => {
+                const color = sketchColorToRgba(item.value);
+                const hex = rgbaToHex(color);
+                const alphaPercent = `${Math.round(color.a * 100)}%`;
+                const name = item.name.split('/').pop() || '';
```
dabbott, line 109: Can use `basename` here I think

packages/app/src/components/theme/SwatchesGrid.tsx
```diff
@@ -1,55 +1,136 @@
...
   );
 
   return (
     <GridView.Root onClick={() => onSelectSwatch(undefined, 'replace')}>
-      {sortedSwatches.map((item) => {
-        const color = sketchColorToRgba(item.value);
-        const hex = rgbaToHex(color);
-        const alphaPercent = `${Math.round(color.a * 100)}%`;
+      {flatSwatchGroup.map((group, index) => {
+        const sortedSwatches = sortBy(group.swatches, 'name');
```
dabbott, line 99: I think these should already be sorted, cause we sorted by 'path' above already

packages/app/src/containers/SwatchesGroups.tsx
```diff
@@ -0,0 +1,97 @@
...
+}));
+
+type SwatchTitles = {
+  name: string;
+  path: string;
+  depth: number;
+};
+
+function flatten(
+  swatch: SwatchGroup,
```
dabbott, line 31: Rename `swatch` => `SwatchGroup`

packages/app/src/containers/SwatchesGroups.tsx
```diff
@@ -0,0 +1,97 @@
...
+  parent: string,
+  depth: number,
+): SwatchTitles[] {
+  const path = (parent ? parent + '/' : '') + swatch.name;
+
+  return Array.prototype.concat.apply(
```
dabbott, line 37: Same as comment above, we can use array spread instead of `concat.apply`

packages/noya-state/src/reducers/application.ts
```diff
@@ -1148,6 +1186,7 @@ export function createInitialState(sketch: SketchFile): ApplicationState {
     selectedObjects: [],
     selectedSwatchIds: [],
     selectedLayerStyleIds: [],
+    selectedGroupSwatch: '',
```
dabbott, line 1189: Rename `selectedGroupSwatch` => `selectedSwatchGroup`

packages/app/src/utils/createSwatchTree.ts
```diff
@@ -0,0 +1,44 @@
...
+    name: '',
+    swatches: [],
+    children: [],
+  };
+
+  function getGroup(pathComponents: string[]): SwatchGroup {
+    let group = root;
+    while (pathComponents.length > 0) {
+      const component = pathComponents.shift()!;
+      if (!group.children) continue;
```
dabbott, line 20: Do we need to check `!group.children`?

packages/noya-state/src/reducers/application.ts
```diff
@@ -1114,6 +1123,35 @@ export function reducer(
...
+      return produce(state, (state) => {
+        const array = state.sketch.document.sharedSwatches?.objects ?? [];
+
+        array.forEach((object: Sketch.Swatch) => {
+          if (!ids.includes(object.do_objectID)) return;
+
+          const prevGroup = dirname(object.name);
+          const name = basename(object.name);
+
+          const group = value
```
dabbott, line 1147: It looks like this code should join the different path segments only if they exist, but I find it a little hard to follow. Will this work?

```ts
const newName = [prevGroup, value, name].filter(x => !!x).join('/')
object.name = newName
```

Possibly a future util, `delimitedPath.join`

dabbott (changes requested):
This is looking good! Mostly some small stuff now.

packages/app/src/components/theme/SwatchesGrid.tsx
milg15: It's not working. I need to use the .apply, maybe there is another way it gives me a typescript error

milg15: It gives me a typescript error 🤔

packages/app/src/components/inspector/NameInspector.tsx
milg15: Oh! It took me a second to understand but I got it, the action uses ids to change the name so i don't need the full name

milg15:
Other than the ts error I mentioned I added all the changes.

packages/app/src/components/theme/SwatchesGrid.tsx
dabbott: What if we do `flatMap` instead of `map`?

dabbott:
👍  I think I got the types wrong, but if we do a `flatMap` it should work

packages/app/src/components/theme/SwatchesGrid.tsx
```diff
@@ -1,57 +1,136 @@
...
 import { SelectionType } from 'noya-state';
-import { memo, useMemo } from 'react';
+import { memo, useMemo, useCallback, Fragment } from 'react';
+import { SwatchGroup, createSwatchTree } from '../../utils/createSwatchTree';
+import { sortBy } from 'noya-utils';
 import ColorSwatch from './ColorSwatch';
+import { delimitedPath } from 'noya-utils';
+
+function flatten(
+  SwatchGroup: SwatchGroup,
```
dabbott, line 13: Argument names should usually be lowercase, `swatchGroup`, but I'll change it another time

packages/noya-utils/src/__tests__/sortBy.test.ts
```diff
@@ -0,0 +1,5 @@
+import { sortBy } from '../index';
+
+test('empty array', () => {
+  expect(sortBy([], '')).toEqual([]);
```
dabbott, line 4: I'm going to merge this, but I'll create a follow-up issue to add some more tests for `sortBy`

packages/noya-utils/src/sortBy.ts
```diff
@@ -0,0 +1,7 @@
+export const sortBy = (array: any[], key: string) =>
```
dabbott, line 1: I'm also going to try to improve the `any` type here

dabbott (approved):

packages/app/src/components/theme/SwatchesGrid.tsx
milg15: oh I'll keep it in mid !

PR #30, opened 2021-04-18
Add generic theme tree utils

dabbott:
Previously the "theme tree" was swatches-specific. This change makes it generic so we can use it with layer styles, text styles, and symbols too.

We also now share the `flatten` function between the swatches grid and inspector.


milg15 (approved):
It looks great! And for what I've tested so far it works with no problem 😄 I'll be using it for the theme styles.

dabbott:
Awesome, thanks for testing 😄

Issue #31, opened 2021-04-18
Add tests for sortBy util


dabbott:
Currently we only test sorting an empty array. Let's add some tests for arrays with items.

PR #35, opened 2021-04-19
More test for sortby

milg15:
Tests for either key, when the values are the same or there is an empty space. Is there another case?

(Solves #31)

### DIFF
--- a/packages/noya-utils/src/__tests__/sortBy.test.ts
+++ b/packages/noya-utils/src/__tests__/sortBy.test.ts
@@ -3,3 +3,83 @@ import { sortBy } from '../index';
 test('empty array', () => {
   expect(sortBy([], '')).toEqual([]);
 });
+
+test('first key', () => {
+  expect(
+    sortBy(
+      [
+        { name: 'Liam', last: 'Smith' },
+        { name: 'Noah', last: 'Benjamin' },
+        { name: 'Sophia', last: 'Brown' },
+        { name: 'Emma', last: 'Jones' },
+        { name: 'Ava', last: 'Davis' },
+      ],
+      'name',
+    ),
+  ).toEqual([
+    { name: 'Ava', last: 'Davis' },
+    { name: 'Emma', last: 'Jones' },
+    { name: 'Liam', last: 'Smith' },
+    { name: 'Noah', last: 'Benjamin' },
+    { name: 'Sophia', last: 'Brown' },
+  ]);
+});
+
+test('second key', () => {
+  expect(
+    sortBy(
+      [
+        { name: 'Liam', last: 'Smith' },
+        { name: 'Noah', last: 'Benjamin' },
+        { name: 'Sophia', last: 'Brown' },
+        { name: 'Emma', last: 'Jones' },
+        { name: 'Ava', last: 'Davis' },
+      ],
+      'last',
+    ),
+  ).toEqual([
+    { name: 'Noah', last: 'Benjamin' },
+    { name: 'Sophia', last: 'Brown' },
+    { name: 'Ava', last: 'Davis' },
+    { name: 'Emma', last: 'Jones' },
+    { name: 'Liam', last: 'Smith' },
+  ]);
+});
+
+test('same value', () => {
+  expect(
+    sortBy(
+      [
+        { name: 'Liam', last: 'Smith' },
+        { name: 'Liam', last: 'Brown' },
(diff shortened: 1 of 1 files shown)


packages/noya-utils/src/__tests__/sortBy.test.ts
```diff
@@ -3,3 +3,83 @@ import { sortBy } from '../index';
 test('empty array', () => {
   expect(sortBy([], '')).toEqual([]);
 });
+
+test('first key', () => {
+  expect(
+    sortBy(
+      [
```
dabbott, line 10: If this is an array we use multiple times, we can move it outside of the `test` function to reuse it.

packages/noya-utils/src/__tests__/sortBy.test.ts
```diff
@@ -3,3 +3,83 @@ import { sortBy } from '../index';
...
+    { name: 'Emma', last: 'Jones' },
+    { name: 'Liam', last: 'Smith' },
+  ]);
+});
+
+test('same value', () => {
+  expect(
+    sortBy(
+      [
+        { name: 'Liam', last: 'Smith' },
```
dabbott, line 53: At this point we've already tested different keys, so we can probably remove `last` to keep it simpler

packages/noya-utils/src/__tests__/sortBy.test.ts
```diff
@@ -3,3 +3,83 @@ import { sortBy } from '../index';
...
+    { name: 'Liam', last: 'Smith' },
+    { name: 'Liam', last: 'Brown' },
+  ]);
+});
+
+test('empty space', () => {
+  expect(
+    sortBy(
+      [
+        { group: 'cloud', ammount: 9 },
```
dabbott, line 70: I think we can remove `ammount` here. And shorten the array to just 2 items, one with an empty string and one without

dabbott (changes requested):
Let's add

- A test that clarifies how we handle uppercase and lowercase comparison, e.g. ["A", "a"]
- A test that contains some "/" since we compare paths with this, e.g. ["a", "aa/b", "a/b/c", "a/b"]

For both of these we probably only need a single a key in the objects we compare, to keep it short

packages/noya-utils/src/__tests__/sortBy.test.ts
dabbott: Actually I guess we wouldn't be able to tell them apart without `last`...

milg15: so I keep them both? or only the first one?

dabbott: Let's keep it the way it is now

packages/noya-utils/src/__tests__/sortBy.test.ts
milg15: where would we place this array?

dabbott: Outside of every function, so e.g.

```ts
const names = [{ name: 'Liam' }, ...]

test('first key', () => {
  expect(sortBy(names, 'name'))...
})

test('second key', () => {
  expect(sortBy(names, 'last'))...
})
```

packages/noya-utils/src/__tests__/sortBy.test.ts
```diff
@@ -3,3 +3,89 @@ import { sortBy } from '../index';
...
+    { letter: 'A' },
+    { letter: 'B' },
+    { letter: 'b' },
+  ]);
+});
+
+test('"/" usage', () => {
+  expect(
+    sortBy(
+      [{ group: 'B/a' }, { group: 'A/d' }, { group: 'B/c' }, { group: 'A/n' }],
```
dabbott, line 82: In this case I want to test comparison between letters, like "B", and "/". Right now the "/" is always in the same position, so not testing that. E.g. we should have "A/BB" and "AA/B" or something.

(Thinking about this some more, we probably can't sort lists of paths using this method way we do in one of the grid components, but we can come back to that another time)

dabbott (changes requested):

packages/noya-utils/src/__tests__/sortBy.test.ts
milg15: Got it!

dabbott (approved):

Issue #32, opened 2021-04-18
Add grouping support for layer styles


dabbott:
We should support all the same grouping features for layer styles. This includes showing a TreeView in the left sidebar for layer styles.

I think it's fine if we mostly duplicate the code between these. If you see any good opportunities for creating shared utils, that could be useful too though, before we add text styles and symbols.

> This relies on https://github.com/noya-app/noya/pull/30, so take a look at that PR first and I'll merge it before starting this

milg15:
I change the files with stylelayer to themelayer, since that's the name we are using in all the actions 😅

PR #38, opened 2021-04-19
support  grouping for layer styles

milg15:
For the most part it was a copy paste with name changes. I think the groups style selectors in the `ApplicationState` could be in an object, instead of creating one for each?

I had to add a small change to the group action because with the new function it wasn't ungrouping. 

Solves #32


dabbott (approved):
This looks good!

> I think the groups style selectors in the ApplicationState could be in an object, instead of creating one for each?

I'm not sure I understand - can you link/screenshot to the part of this code this is?

milg15:
I'm currently handling the groups like this but I I'll have to add more for the other components. I was thinking this could be and object instead of one for each.  

image

milg15:
I'll merge this PR after you approve this PR #39 😄

milg15:
The moment of the truth 🤣 Will it merge? It seems like it

PR #39, opened 2021-04-19
Duplicate in the context menu

milg15:
Duplicate with in the context menu. Pending: Duplicate theme style


milg15:
I will prefer to merge this branch into the groupthemestyles and then that one to main, so it's easier to solve any merge issues 😄

packages/noya-state/src/reducers/application.ts
```diff
@@ -134,6 +134,8 @@ export type Action =
       sharedStyleId: string,
       style: Sketch.Style | undefined,
     ]
+  | [type: 'duplicateColorSwatch', id: string[]]
```
dabbott, line 137: I think we mostly call them `Swatch` and not `ColorSwatch` in actions, e.g. `setSwatchName`, `removeSwatch`, `groupSwatches`. So let's call this `duplicateSwatch`.

packages/noya-utils/src/getIncrementedName.ts
```diff
@@ -0,0 +1,12 @@
+const numberSuffixRegExp = /(\s\d+)$/;
+
+export function getIncrementedName(name: string): string {
```
dabbott, line 3: What you have works and has all the right logic, but I think we can simplify it by including the "prefix" (non-number) part of the string in the regex. Does this work?

```ts
const numberSuffixRegExp = /(.*?)(\s\d+)?$/
  
function getIncrementedName(name: string): string {
  const [, prefix, number] = name.match(numberSuffixRegExp)
  const parsedNumber = (number ? parseInt(number) : 1)
  return `${prefix} ${parsedNumber + 1}`
}
```

dabbott (changes requested):
We also need to figure out the issue you mentioned around duplicating "Test" twice, to get "Test 2" and "Test 3"

packages/noya-utils/src/getIncrementedName.ts
milg15: This is the result of using that regexp 😅 
image

milg15: Oh I just applied the full function and it works. I'll figure out the other part for the name 🤔 I think I Can use a list of the names that are the same

milg15:
I added a function to handle the name problem I mentioned but I think this function I created could be move somewhere else

packages/noya-state/src/reducers/application.ts
```diff
@@ -875,6 +882,77 @@ export function reducer(
...
+          _class: 'swatchContainer',
+          do_objectID: uuid(),
+          objects: [],
+        };
+
+        sharedSwatches.objects.forEach((swatch: Sketch.Swatch) => {
+          if (!ids.includes(swatch.do_objectID)) return;
+          const swatchColor = swatch.value;
+
+          const lastName = sharedSwatches.objects
```
dabbott, line 900: This will work if the names are in the correct order, but if the user rearranges the swatches, there may be a larger number that comes first: `["Test 3", "Test 2", "Test"]`. Since I'm hoping to use this same function for layers in the future, I think we want to handle that case.

Let's update `getIncrementedName` to also take the array of other names. Maybe something like:

```ts
function getIncrementedName(originalName: string, names: string[] = []): string {
  // ...
  const numbers = [originalName, ...names]
    .filter(name => name.startsWith(prefix))
    .map(name => /* map to just the parsed number */)
    .sort()
  const nextNumber = numbers.length > 0 ? numbers[numbers.length - 1] : 1
  // ...
}
```

Then to use it here:

```ts
const names = sharedSwatches.objects.map(s => s.name)
const newName = getIncrementedName(swatch.name, names)
```

dabbott (changes requested):

packages/noya-state/src/reducers/application.ts
milg15: yes I was also thinking this could be handle in the same function!

milg15:
I was thinking the function could also have a type similar to sortBy where the value in originalName has to be in names. I don't know if this is possible or necessary. I tried to do it but it didn't work 😆

dabbott:
I was thinking of combining originalName and names too! But we wouldn't know what "prefix" to use, so I'm not sure it'd work. We could require the original name to be the first item in the names array, but I think it's clearer separate.

milg15:
I'll fix the conflicting file now so I can merge them.

packages/noya-utils/src/getIncrementedName.ts
```diff
@@ -0,0 +1,19 @@
...
+  originalName: string,
+  names: string[],
+): string {
+  const [, prefix] = originalName.match(numberSuffixRegExp) || [];
+
+  const numbers = [originalName, ...names]
+    .filter((name) => name.startsWith(prefix))
+    .map((name) => {
+      const [, , number] = name.match(numberSuffixRegExp) || [];
+      return parseInt(number || '1');
```
dabbott, line 13: This should work but I think it'd be clearer as `return number ? parseInt(number) : 1`, since there isn't really a need to parse '1' as a string

dabbott (approved):
Looks good! I think there are other cases we might want to test for getIncrementedName, but that can happen another time

PR #42, opened 2021-04-21
Correction: Show the theme inspector in the layers inspectors

milg15:
I changed the files names and it wasn't showing the inspector in the layers inspectors

Before: 
image

Now:
image


dabbott (commented):
Is the file `ThemeStyleInspector` still used or should we delete it?

milg15:
> Is the file `ThemeStyleInspector.tsx` still used or should we delete it?

Yes!! But it doesn't something different! That file has the inspector for the ThemeStyles in the ThemeWindow

dabbott (approved):

PR #44, opened 2021-04-23
UI for the text style inspector

milg15:
The changes in the `inspector.ts` are for testing. I have some states for testing in the file :D and the **Auto Height** label should be smaller.
image


packages/app/src/containers/Inspector.tsx
```diff
@@ -51,15 +52,29 @@ export default memo(function Inspector() {
...
           {...dimensionsInspectorProps}
           onSetRotation={handleSetRotation}
         />
         <Spacer.Vertical size={10} />
-      </Fragment>,
+      </Fragment>
+    );
+
+    // Only show the text inspector if layer is a text
+    if (selectedLayers.length === 1 && selectedLayers[0]._class === 'text') {
```
dabbott, line 67: Let's put the `<TextStyleInspector />` in the same `views` array as all the other inspectors. It wasn't clear from the screenshot/video, cause I collapsed the other sections, but text layers can also have fills, opacity, etc.

The logic Sketch uses is: if at least one text layer is selected, then show the TextStyleInspector. Even if it's 1 text layer and 1 rectangle selected, we still show the TextStyleInspector. E.g. `hasTextLayers && <TextStyleInspector>`

Screen Shot 2021-04-23 at 9 40 41 AM

packages/app/src/components/inspector/TextStyleInspector.tsx
```diff
@@ -0,0 +1,377 @@
...
+  onChangeFontVerticalAlignment,
+}: TextAligmentRowProps) {
+  return (
+    <InspectorPrimitives.Section>
+      <InspectorPrimitives.SectionHeader>
+        <InspectorPrimitives.Title>Alignment</InspectorPrimitives.Title>
+      </InspectorPrimitives.SectionHeader>
+      <Spacer.Vertical size={10} />
+      <InspectorPrimitives.Row>
+        <LabeledElementView renderLabel={() => 'Auto Height'}>
```
dabbott, line 179: I think if we make it a `Label.Label` it should be the right size?

```tsx
renderLabel={() => <Label.Label>Auto Height</Label.Label>}
```

packages/app/src/components/inspector/TextStyleInspector.tsx
```diff
@@ -0,0 +1,377 @@
...
+
+const TextStyleRow = memo(function TextStyleRow({
+  fontColor,
+  fontFamily,
+  fontWeight,
+  onChangeFontColor,
+  onChangeFontFamily,
+  onChangeFontWeight,
+}: TextStyleRowProps) {
+  const id = '1';
```
dabbott, line 69: This number `id` was only need for fills/borders/shadows because they're used in an array. We can delete this

packages/app/src/components/inspector/TextStyleInspector.tsx
```diff
@@ -0,0 +1,377 @@
...
+    </InspectorPrimitives.Section>
+  );
+});
+const TextOptionsRow = memo(function TextOptionsRow({
+  fontCase,
+  fontDecorator,
+  onChangeFontCase,
+  onChangeFontDecorator,
+}: TextOptionsRowProps) {
+  const id = '1';
```
dabbott, line 247: We can remove this `id` too

packages/app/src/components/inspector/TextStyleInspector.tsx
```diff
@@ -0,0 +1,377 @@
...
+  onChangeFontFamily: (value: string) => void;
+  onChangeFontWeight: (value: string) => void;
+  onChangeFontColor: (color: Sketch.Color) => void;
+}
+
+interface TextAligmentRowProps {
+  fontAlignment: string;
+  fontHorizontalAlignment: string;
+  fontVerticalAlignment: string;
+  onChangeFontAlingment: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
dabbott, line 45: Spelling: `onChangeFontAlingment` => `onChangeFontAlignment`

packages/app/src/components/inspector/TextStyleInspector.tsx
```diff
@@ -0,0 +1,377 @@
...
+          value={fontHorizontalAlignment}
+          onValueChange={onChangeFontHorizontalAlignment}
+        >
+          <RadioGroup.Item value="left" tooltip="Left">
+            <TextAlignLeftIcon />
+          </RadioGroup.Item>
+          <RadioGroup.Item value="center" tooltip="Center">
+            <TextAlignCenterIcon />
+          </RadioGroup.Item>
+          <RadioGroup.Item value="Right" tooltip="Right">
```
dabbott, line 211: Lowercase value for consistency, `value="right"`

packages/app/src/components/inspector/TextStyleInspector.tsx
```diff
@@ -0,0 +1,377 @@
...
+    },
+    [decoratorInputId, transformInputId],
+  );
+  return (
+    <InspectorPrimitives.Section>
+      <InspectorPrimitives.SectionHeader>
+        <InspectorPrimitives.Title>Text Options</InspectorPrimitives.Title>
+      </InspectorPrimitives.SectionHeader>
+      <Spacer.Vertical size={10} />
+      <InspectorPrimitives.Row id={'id'}>
```
dabbott, line 271: I'm not sure if this id is needed, but it doesn't hurt anything either. The ids are mainly for drag n drop or label stuff. If we keep it, it should be a different string, like `text-options` or something though

dabbott (changes requested):
This looks great! We can leave the hardcoded `useState` data for now, but we'll replace it soon

packages/app/src/components/inspector/TextStyleInspector.tsx
milg15: oh! make sense 😅

packages/app/src/components/inspector/TextStyleInspector.tsx
milg15: I'll remove it in this case

packages/app/src/containers/Inspector.tsx
```diff
@@ -51,6 +52,10 @@ export default memo(function Inspector() {
             rotation: undefined,
           };
 
+    const hasTextLayer =
+      selectedLayers.length > 0 &&
```
dabbott, line 56: The `.some` should handle the empty array case correctly already, so we shouldn't need `selectedLayers.length > 0`

dabbott (approved):
Looks good!

I think you should be able to make the small `.length` change and the PR will still be "approved", so you can merge when you're done

PR #46, opened 2021-04-23
rename produce state to draft

souporserious:
This renames all of the `produce` functions from immer to use `draft` instead of `state` so it's easier to track the differences.


dabbott (approved):
Looks good! Thanks for doing this. I'll go ahead and merge it now

PR #48, opened 2021-04-25
Text style actions

### LINKED ISSUE #45: Support modifying text styles
## Feature

We should allow displaying and modifying the styles of text layers. This means populating the input fields in the `TextStyleInspector` and creating actions to modify the values in the `state.sketch` file data. We can use `TextLayers.sketch` to test the various style properties.

For reference, this is how `TextLayers.sketch` looks currently:

| App | Screenshot |
| --- | --- |
| Sketch | Screen Shot 2021-04-23 at 3 11 00 PM |
| Noya | Screen Shot 2021-04-23 at 3 11 29 PM |

## Step 1

First, I want to document how the JSON looks for these text styles. Some of it we can figure out from the TypeScript definitions for the Sketch file format, but some of it isn't documented there. I'd like to have a short description/example of where each style property is stored (we can write it in this GitHub issue).

## Step 2

Once we've documented how things work, we can implement it.

## Not part of this issue

- Modifying the actual text/string of the layer (and not just the style) will be complex and will come later.
- Sometimes the layout of the text will be a tiny bit off. E.g. one character might wrap when it's not supposed to. That's OK for now
Screen Shot 2021-04-23 at 3 21 21 PM
- Some properties might be hard to implement at the moment. For example, font family will require loading font files from... somewhere. Anything difficult we can leave for another time

milg15:
TODO: 

- [x] Handle multiple texts (show the multiple label if different a the right value if all share the value)
- [x] Change multiple texts
- [x] Handle mixed styles in one text 
- [x]  Maybe add some new selectors for repeating actions for text 🤔 
- [ ] Call the fontFamilies and fontCase information array from somewhere else

closes #45


packages/app/src/components/inspector/TextAlignmentRow.tsx
```diff
@@ -0,0 +1,109 @@
...
+  PinTopIcon,
+  AlignCenterVerticallyIcon,
+} from '@radix-ui/react-icons';
+import { Label, RadioGroup, LabeledElementView } from 'noya-designsystem';
+import { Spacer } from 'noya-designsystem';
+import { memo } from 'react';
+import * as InspectorPrimitives from './InspectorPrimitives';
+
+interface TextAligmentRowProps {
+  fontAlignment: number;
```
dabbott, line 19: For naming, let's call these:

`textLayout`
`textHorizontalAlignment`
`textVerticalAlignment`

packages/app/src/containers/TextStyleInspector.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const [, dispatch] = useApplicationState();
+
+  //Todo: Handle multiple texts
+  const selectedTextStyles = useShallowArray(
+    useSelector(Selectors.getSelectedTextStyles),
+  );
+
+  const seletedText = useShallowArray(useSelector(Selectors.getSelectedText));
+
+  const fontColor = useMemo(
```
dabbott, line 29: I think we can combine most of these into a single `useMemo`. When any property of the text style changes, we're going to have to re-render almost everything anyway, so it shouldn't make a difference in terms of performance.

Something like the `getTextStyleAttributes` function I added to `SketchText.tsx`

packages/app/src/components/inspector/TextStyleRow.tsx
```diff
@@ -0,0 +1,171 @@
...
+  onChangeLineSpacing,
+  onChangeLetterSpacing,
+  onChagenParagraphSpacing,
+}: TextStyleRowProps) {
+  const characterInputId = `char`;
+  const lineInputId = `line`;
+  const paragraphInputId = `paragraph`;
+
+  const [family, size] = useMemo(
+    () => fontFamily.replace('MT', '').split('-'),
```
dabbott, line 49: FYI, we'll have to get font weight (light, bold, etc) from the font file itself I think (the .ttf or .otf). Using the name isn't reliable. We don't have a way to do that yet though, so you can leave this for now.

packages/noya-state/src/reducers/application.ts
```diff
@@ -169,6 +169,17 @@ export type Action =
       swatchId: string | string[],
       name: string | undefined,
     ]
+  | [type: 'setTextColor', value: Sketch.Color]
```
dabbott, line 172: Maybe we can move all of these into a `reducers/textStyles.ts`, similar to `reducers/style.ts`? The main difference would be, the `state` we pass in would need to be the `layer` I think.

packages/noya-state/src/reducers/application.ts
```diff
@@ -1257,6 +1277,231 @@ export function reducer(
...
+      const [, color] = action;
+
+      const pageIndex = getCurrentPageIndex(state);
+      const layerIndexPaths = getSelectedLayerIndexPaths(state);
+
+      return produce(state, (draft) => {
+        accessPageLayers(draft, pageIndex, layerIndexPaths).forEach((layer) => {
+          if (layer._class !== 'text') return;
+          const encoded = layer.style?.textStyle?.encodedAttributes;
+          const attributes = layer.attributedString.attributes[0].attributes;
```
dabbott, line 1290: Would we want to do a `forEach` on `attributedString.attributes` to change all of them? (same for the other actions)

packages/noya-state/src/reducers/application.ts
```diff
@@ -1257,6 +1277,231 @@ export function reducer(
...
+
+          if (encoded) {
```
dabbott, line 1292: Maybe both the `encoded` and `attributes` case here could call a `changeColor` function or something, if they're similar.

packages/noya-state/src/reducers/application.ts
```diff
@@ -1257,6 +1277,231 @@ export function reducer(
...
+            if (!encoded.MSAttributedStringColorAttribute) {
```
dabbott, line 1293: Can we always assign a new color?

```ts
encoded.MSAttributedStringColorAttribute = color;
```

What does the case where we assign each r/g/b/a value individually do?

packages/noya-state/src/reducers/application.ts
```diff
@@ -1257,6 +1277,231 @@ export function reducer(
...
+          const encoded =
+            layer.style?.textStyle?.encodedAttributes
+              .MSAttributedStringFontAttribute.attributes;
+          const attributes =
+            layer.attributedString.attributes[0].attributes
+              .MSAttributedStringFontAttribute.attributes;
+
+          if (!encoded) return;
+
+          const face = encoded.name.split('-')[1]
```
dabbott, line 1335: We can leave this logic, but don't worry about getting it perfect. We'll need more info from the font file before we can do this 100% correctly

packages/noya-state/src/reducers/application.ts
```diff
@@ -1257,6 +1277,231 @@ export function reducer(
...
+      const [, value] = action;
+
+      const pageIndex = getCurrentPageIndex(state);
+      const layerIndexPaths = getSelectedLayerIndexPaths(state);
+
+      return produce(state, (draft) => {
+        accessPageLayers(draft, pageIndex, layerIndexPaths).forEach((layer) => {
+          if (layer._class !== 'text') return;
+          const encoded = layer.style?.textStyle?.encodedAttributes;
+          const attributes = layer.attributedString.attributes[0].attributes;
```
dabbott, line 1381: I wonder if we could do something like (and this might not work, due to type differences):

```ts
const allAttributes = [encoded, ...layer.attributedString.attributes]

allAttributes.forEach((attribute) => {
  switch (action[0]) {
    case 'setTextLetterSpacing': {
      attribute.kerning = value;
      break;
    }
  }
})
```

packages/app/src/containers/TextStyleInspector.tsx
```diff
@@ -0,0 +1,220 @@
...
+        green: 0.5,
+        alpha: 0.5,
+      } as Sketch.Color),
+    [fontColor],
+  );
+
+  const lineSpacing = useMemo(
+    () =>
+      selectedTextStyles.map(
+        (style) => style?.textStyle?.encodedAttributes.kerning,
```
dabbott, line 74: These 2 properties are flipped: `kerning` should be `letterSpacing` and `maximumLineHeight` should be `lineSpacing`

dabbott (commented):
This looks good! I think you're right that there are some ways we can clean things up, and I left a few ideas.

I added a commit that supports previewing a few more properties directly to this branch, so make sure to pull!

packages/app/src/components/inspector/TextStyleRow.tsx
milg15: Yeah, I did noticed that the MT wasn't in all the files 😅 But I also have all the fonts in an array, I'm guessing this will be handle better later. It's good to keep in mind.

packages/noya-state/src/reducers/application.ts
milg15: Yes! I'll add them into an unique file !

packages/noya-state/src/reducers/application.ts
milg15: It was to not change the swatchId  when you change the color but now I remember it should change it 😅

packages/noya-state/src/reducers/application.ts
milg15: I think I'll have to have add an index 🤔 to handle those attributes, because I am guessing the other attributes are for different areas of the text. I will add the foreach and later the conditionals for to access the right string

packages/noya-state/src/reducers/application.ts
milg15: I added it but did it in a way

packages/noya-state/src/reducers/application.ts
milg15: I did something like it but I'm not sure if it's optimal I added the "I am not sure if it's optimal comment"

packages/app/src/containers/TextStyleInspector.tsx
milg15: Oh that's a cool and cleaner solution 👍

packages/noya-state/src/reducers/application.ts
```diff
@@ -1257,6 +1268,43 @@ export function reducer(
...
+        accessPageLayers(draft, pageIndex, layerIndexPaths).forEach((layer) => {
+          if (layer._class !== 'text' || layer.style?.textStyle === undefined)
+            return;
+
+          if (action[0] === 'setTextAlignment') {
+            layer.textBehaviour = action[1];
+            return;
+          }
+          // It's this optimal ?
+          layer.style.textStyle = textStyleReducer(
```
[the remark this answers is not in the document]
milg15: They look different encodedAttributes and attributes, they don't share all the properties I think so it gave an error 🤔

packages/app/src/components/inspector/TextLayoutRow.tsx
```diff
@@ -0,0 +1,109 @@
...
+  PinTopIcon,
+  AlignCenterVerticallyIcon,
+} from '@radix-ui/react-icons';
+import { Label, RadioGroup, LabeledElementView } from 'noya-designsystem';
+import { Spacer } from 'noya-designsystem';
+import { memo } from 'react';
+import * as InspectorPrimitives from './InspectorPrimitives';
+
+interface TextLayoutRowProps {
+  textLayout: number;
```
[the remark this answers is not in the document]
milg15: So I don't use the event as the parameter anymore ? 

``` onChangeTextAlignment: (event: React.ChangeEvent<HTMLInputElement>) => void;``` ?

File: packages/noya-state/src/reducers/application.ts [hunk 13]
```diff
@@ -1002,6 +1004,15 @@ export function reducer(
             Layers.visit(page, (layer) => {
               if (!layer.style) return;
               visitColors(layer.style, changeColor);
+
+              if (layer._class !== 'text') return;
```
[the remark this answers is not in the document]
milg15: 👍

packages/app/src/components/inspector/TextLayoutRow.tsx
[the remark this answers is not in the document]
dabbott: Yeah, I think it's better to keep the event handling contained within this component, since the parent only needs to know the value of the input

packages/noya-state/src/reducers/application.ts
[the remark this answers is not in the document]
dabbott: Can we do it just for the properties they both share? We might need to make a union of both types, and use a generic to make sure we get the correct type out. Something like:

```ts
type CommonStringAttributes = Sketch.StringAttribute['attributes'] | Sketch.TextStyle['encodedAttributes']

function stringAttributeReducer<T extends CommonStringAttributes>(state: T, action: Action): T {
  state.kerning = 123
  return state
}
```

milg15:
I think the code in `TextStyleInspector.tsx` sucks 😆 but for now it does the work. I was wondering if it was better to group it, separarte it or maybe there is a function I could create to work better.

packages/app/src/containers/TextStyleInspector.tsx
```diff
@@ -0,0 +1,257 @@
...
+            dispatch('setTextCase', parseInt(event.target.value));
+          },
+          [dispatch],
+        )}
+      />
+    </>
+  );
+});
+
+/**
```
dabbott, line 253: Remove comment

packages/noya-state/src/reducers/application.ts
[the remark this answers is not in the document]
dabbott: We should still try this

packages/noya-state/src/selectors.ts
```diff
@@ -134,6 +134,20 @@ export const getSelectedStyles = (state: ApplicationState): Sketch.Style[] => {
...
+  state: ApplicationState,
+): Sketch.Style[] => {
+  return getSelectedLayers(state)
+    .filter((layer) => layer._class === 'text')
+    .flatMap((layer) => (layer.style ? [layer.style] : []));
+};
+
+export const getSelectedText = (state: ApplicationState): Sketch.Text[] => {
+  return getSelectedLayers(state)
+    .filter((layer) => layer._class === 'text')
```
dabbott, line 147: We can use a type guard here to avoid the map:

```ts
return getSelectedLayers(state)
  .filter((layer): layer is Sketch.Text => layer._class === 'text')
```

dabbott: And rename to `getSelectedTextLayers`

packages/noya-state/src/selectors.ts
```diff
@@ -134,6 +134,20 @@ export const getSelectedStyles = (state: ApplicationState): Sketch.Style[] => {
     : getSelectedLayerStyles(state).map((style) => style.value);
 };
 
+export const getSelectedTextStyles = (
```
dabbott, line 137: This can call `getSelectedTextLayers` rather than filtering within this function

dabbott (changes requested):
This is looking good! I think the features are all here 💯. There are still a few code organization things I want to try to improve before we merge it

packages/app/src/components/inspector/TextOptionsRow.tsx
```diff
@@ -0,0 +1,114 @@
...
+  Label,
+  Select,
+  RadioGroup,
+  LabeledElementView,
+} from 'noya-designsystem';
+import { Spacer } from 'noya-designsystem';
+import { useCallback, useMemo, memo } from 'react';
+import * as InspectorPrimitives from './InspectorPrimitives';
+
+export enum SimpleTextDecoration {
```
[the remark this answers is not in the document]
milg15: okey!

packages/app/src/containers/TextStyleInspector.tsx
```diff
@@ -0,0 +1,257 @@
...
+        }) &&
+        attributes.every(
+          (a) =>
+            a.attributes.paragraphStyle?.maximumLineHeight ===
+            paragraphStyle?.maximumLineHeight,
+        )
+          ? paragraphStyle?.maximumLineHeight ?? 22
+          : undefined,
+      horizontalAlignment:
+        layers.every((layer) => {
```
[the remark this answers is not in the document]
milg15: I can't do this because we have a different default value in some cases. 

I did created the function but it returns bolean. You can see it now and check if this is a good approach.

packages/noya-state/src/reducers/application.ts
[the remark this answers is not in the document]
milg15: Oh! I don't really understand this 🤔 Do I move the actions that don't relate to this?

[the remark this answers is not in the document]
milg15: ` if (draft._class === 'stringAttribute')` this ones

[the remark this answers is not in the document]
dabbott: Yeah, I think we should try moving the actions that apply to both `StringAttribute` and `TextStyle` objects into a new `stringAttributeReducer`, and the actions that only apply to one or the other can go in the main application reducer (for now)

dabbott (changes requested):
This is very close! I think the only thing left is to try creating a `stringAttributeReducer`. If it's still not clear what to do, we can also merge this now and I can try it.

The `getMultiValue` part could possibly have some improvements, but we can leave it like this until we implement text editing

packages/noya-state/src/reducers/application.ts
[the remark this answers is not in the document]
milg15: Ohhh! Okey I think I got it. I'll change it then

milg15:
I hope I understood it this time

dabbott (approved):
This is close! I'm probably not describing the change very well. I'm going to go ahead and merge this, and I'll update it another time

PR #52, opened 2021-05-06
Create Text Theme Styles

milg15:
I added some preview with css and joined some actions in the application

### DIFF
--- a/packages/app/src/components/theme/TextStyle.tsx
+++ b/packages/app/src/components/theme/TextStyle.tsx
@@ -5,7 +5,7 @@ import { sketchColorToRgbaString, sketchColorToRgba } from 'noya-designsystem';
 
 // It has very basic bacground color preview.
 const TextStylePrev = styled.span<{ size: number; backgroundColor: string }>(
-  ({ theme, color, size, backgroundColor }) => ({
+  ({ color, size, backgroundColor }) => ({
     fontSize: size,
     backgroundColor: backgroundColor,
     height: '90%',
@@ -14,6 +14,8 @@ const TextStylePrev = styled.span<{ size: number; backgroundColor: string }>(
     justifyContent: 'center',
     alignItems: 'center',
     color: color,
+    pointerEvents: 'none',
+    userSelect: 'none',
   }),
 );
 
--- a/packages/app/src/components/theme/TextStyle.tsx
+++ b/packages/app/src/components/theme/TextStyle.tsx
@@ -4,28 +4,52 @@ import styled from 'styled-components';
 import { sketchColorToRgbaString, sketchColorToRgba } from 'noya-designsystem';
 
 // It has very basic bacground color preview.
-const TextStylePrev = styled.span<{ size: number; backgroundColor: string }>(
-  ({ color, size, backgroundColor }) => ({
-    fontSize: size,
-    backgroundColor: backgroundColor,
-    height: '90%',
-    width: '90%',
-    display: 'flex',
-    justifyContent: 'center',
-    alignItems: 'center',
-    color: color,
-    pointerEvents: 'none',
-    userSelect: 'none',
-  }),
-);
+export type SimpleTextDecoration = 'none' | 'underline' | 'strikethrough';
 
+const TextStylePrev = styled.span<{
+  size: number;
+  backgroundColor: string;
+  textTransform?: number;
+  textDecoration: SimpleTextDecoration;
+}>(({ color, size, textTransform, textDecoration, backgroundColor }) => ({
+  fontSize: size,
+  backgroundColor: backgroundColor,
+  height: '90%',
+  width: '90%',
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+  color: color,
+  textTransform:
+    textTransform === 1
+      ? 'uppercase'
+      : textTransform === 2
+      ? 'lowercase'
+      : 'none',
+  pointerEvents: 'none',
+  userSelect: 'none',
+  textDecoration:
+    textDecoration === 'underline'
+      ? 'underline'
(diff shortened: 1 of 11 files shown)


milg15:
I will make another commit where I join the ThemeGroups into one file, but I need some help on where to place the selector for the elements.

packages/app/src/containers/ThemeTextStyleInspector.tsx
```diff
@@ -0,0 +1,236 @@
...
+          (s) =>
+            s?.textStyle?.encodedAttributes
+              ?.MSAttributedStringTextTransformAttribute,
+        ),
+      )
+        ? encodedAttributes?.MSAttributedStringTextTransformAttribute ??
+          Sketch.TextTransform.None
+        : undefined,
+      textDecoration: encodedAttributes?.underlineStyle
+        ? 'underline'
```
dabbott, line 84: If we do `'underline' as const`, `'strikethrough' as const`, ...,  we should be able to avoid the `textDecoration as SimpleTextDecoration` below

packages/app/src/containers/ThemeTextStyleInspector.tsx
```diff
@@ -0,0 +1,236 @@
...
+import TextAlignmentRow from '../components/inspector/TextLayoutRow';
+import TextStyleRow from '../components/inspector/TextStyleRow';
+import getMultiValue from '../utils/getMultiValue';
+
+const TextStyleInspector = memo(function TextStyleInspector() {
+  const [, dispatch] = useApplicationState();
+
+  const seletedText = useShallowArray(useSelector(Selectors.getSelectedStyles));
+
+  const getTextStyleAttributes = useCallback((styles: Sketch.Style[]) => {
```
dabbott, line 29: Is this the selector you mentioned in your comment? 

Maybe as a starting point, we can write a helper function in `noya-state/src/textStyleSelectors.ts` that extracts just the properties shared between the two inspectors? 

The `selectors.ts` file is getting large, so probably I will move it into a directory and split things into separate files, but I want to wait till I see Leanne's PR in case it causes conflicts

dabbott (changes requested):
This is good! I think I left feedback to all of your comments, but let me know if I missed something.

Have you seen this PR? https://github.com/noya-app/noya/pull/51 I thought GitHub would send an email notification, but maybe not... Anyway, we're going to need to merge with `main` to pick up a few changes. There are a few files that overlap, so if you run into trouble I can also try it

milg15:
I didn't see the PR, I'll take a look and do the changes 😄

packages/noya-state/src/reducers/application.ts
```diff
@@ -953,26 +1032,80 @@ export function reducer(
...
+
+          textStyles.objects.push(newSharedStyle);
+        });
+
+        draft.sketch.document.layerTextStyles = textStyles;
+      });
+    }
+    case 'selectSwatch':
+    case 'selectThemeStyle':
+    case 'selectTextStyle': {
```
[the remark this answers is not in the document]
milg15: I added the helper function to the selectors.ts file.

packages/noya-state/src/reducers/application.ts
```diff
@@ -1226,15 +1315,28 @@ export function reducer(
       });
     }
     case 'setSelectedSwatchGroup':
-    case 'setSelectedThemeStyleGroup': {
+    case 'setSelectedThemeStyleGroup':
+    case 'setSelectedTextStyleGroup': {
```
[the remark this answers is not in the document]
milg15: Maybe we can later move this themeComponents actions to a different file

packages/app/src/containers/ThemeTextStyleInspector.tsx
milg15: I created the file 😅  I hope it's not too far of to what you meant. With this selector I was able to combine both Inspector in one.

milg15:
This PR turned out way bigger than I expected 😅

milg15:
I did a rebase because just changing the application.ts gave me errors when testing the file 😅  I am not sure if this was the best approach to solve this but it works.

Touching `application.ts` in different branches it's always leads to weird behavior because it affects different parts of the app..

packages/noya-state/src/reducers/application.ts
```diff
@@ -1272,49 +1388,92 @@ export function reducer(
...
+                  break;
+                }
+                case 'setTextCase': {
+                  const encoded = layer.style?.textStyle?.encodedAttributes;
+                  if (!encoded) return;
+                  encoded.MSAttributedStringTextTransformAttribute = action[1];
+                  break;
+                }
+                default: {
+                  layer.style.textStyle = stringAttributeReducer(
```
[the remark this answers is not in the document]
milg15: I think I was able to add it all. Let me know if I missed something 😄 This PR it's a bit big

packages/app/src/components/theme/TextStyle.tsx
```diff
@@ -0,0 +1,76 @@
+import Sketch from '@sketch-hq/sketch-file-format-ts';
+import { memo, useMemo } from 'react';
+import styled from 'styled-components';
+import { sketchColorToRgbaString, sketchColorToRgba } from 'noya-designsystem';
+
+// It has very basic bacground color preview.
+export type SimpleTextDecoration = 'none' | 'underline' | 'strikethrough';
```
dabbott, line 7: We should only have one copy of this, but I'll fix it after we merge. I'm going to rearrange a few other things too

dabbott (approved):
This looks great! Nice job 🔥

I'm going to merge this and see if I can split `application.ts` into smaller pieces to avoid some of these merge conflicts in the future

PR #55, opened 2021-05-08
Refactor actions and selectors

dabbott:
This PR breaks up the monolithic `selectors.ts` and `application.ts` reducer into more manageable chunks.

@milg15


milg15:
This looks amazing! I will try it myself because I might have to do a change on one of the inspectors because one of them relied on the selectors to show either receive either the text or theme style to modify the fill and all that.

milg15:
As far as I've tested, It all works correctly. 👍

PR #57, opened 2021-05-10
Theme Text Style Inspector

milg15:
Similar to the theme layer inspector.

There is a weird preview bug were the preview and the cahgnes aren't syncing correctly.

### DIFF
--- a/packages/app/src/containers/Inspector.tsx
+++ b/packages/app/src/containers/Inspector.tsx
@@ -1,5 +1,6 @@
 import { Divider, Spacer } from 'noya-designsystem';
 import { Selectors, SetNumberMode } from 'noya-state';
+import * as Layers from 'noya-state/src/layers';
 import { Fragment, memo, useCallback, useMemo } from 'react';
 import DimensionsInspector from '../components/inspector/DimensionsInspector';
 import {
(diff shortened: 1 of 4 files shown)


packages/app/src/containers/Inspector.tsx
```diff
@@ -52,6 +53,7 @@ export default memo(function Inspector() {
           };
 
     const hasTextLayer = selectedLayers.some((l) => l._class === 'text');
+    const hasAllTextLayer = selectedLayers.every((l) => l._class === 'text');
```
dabbott, line 56: I recently added a `Layers.isTextLayer` the we can use instead of the `_class` comparison. Not super important, but looks a little clearer I think

packages/app/src/containers/ThemeTextInspector.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const sharedTextStyles = useShallowArray(
+    useSelector(Selectors.getSharedTextStyles),
+  );
+  const selectedLayers = useShallowArray(
+    useSelector(Selectors.getSelectedLayers),
+  );
+
+  const selectedTextStyleId = useMemo(
+    () =>
+      !selectedLayers.every(
```
dabbott, line 35: We should have a helper for this pattern since we use it so often: if every value in an array is the same, we return that value, otherwise undefined

I originally wanted `getMultiValue` to handle this case, e.g: 

```ts
getMultiValue(selectedLayers.map(v => v.id))
``` 

I remember we changed `getMultiValue` to return a `boolean`... but I'm not sure if the boolean-returning version is really an improvement over just calling `every` directly. Maybe we change `getMultiValue` to return the `T | undefined`, and change the other usages back to `.every`?

dabbott (approved):
Looks great! I noticed a couple small things I think we can improve, but I'll go ahead and approve this now

packages/app/src/containers/ThemeTextInspector.tsx
milg15: I thinking we could have two functions because I do use the getMultiValue with boolean a lot, maybe a different functions.

dabbott: Ok, we can try that. Let's do it in a separate issue like you suggested

PR #61, opened 2021-05-12
Symbol rendering

dabbott:
This adds support for rendering symbol instances and symbol masters. We don't yet handle overrides, scaling, or other operations that can be done on symbols.

Screen Shot 2021-05-12 at 2 11 02 PM

@milg15


milg15 (approved):
It looks great and not that complex

PR #66, opened 2021-05-14
Create Symbols from Inspector

milg15:
Allows to create symbols from the LayerList. It's creates a masterSymbol in a Symbols page and if the page doesn't exist it creates it.

### DIFF
--- a/packages/noya-state/src/reducers/layerReducer.ts
+++ b/packages/noya-state/src/reducers/layerReducer.ts
@@ -320,19 +320,16 @@ export function layerReducer(
         const symbolsPage =
           symbolsPageIndex === -1
             ? createSymbolsPage(draft.sketch.pages, draft.sketch.user)
-            : (Layers.access(
-                draft.sketch.pages[symbolsPageIndex],
-                selectedIndexPaths[0].slice(0, -1),
-              ) as Sketch.Page);
+            : draft.sketch.pages[symbolsPageIndex];
 
-        const crrPage = Layers.access(
+        const crrParent = Layers.access(
           draft.sketch.pages[pageIndex],
           selectedIndexPaths[0].slice(0, -1),
         ) as Sketch.Page;
 
         //ANCHOR: ERROR WHEN ADDING THE SYMBOL IN THE SAME PAGEE
         symbolsPage.layers.push(symbolMasters);
-        crrPage.layers.splice(
+        crrParent.layers.splice(
           lastIndexPath[lastIndexPath.length - 1],
           0,
           symbolInstance,
--- a/packages/noya-state/src/reducers/layerReducer.ts
+++ b/packages/noya-state/src/reducers/layerReducer.ts
@@ -29,42 +29,36 @@ export type LayerAction =
       selectionType?: SelectionType,
     ];
 
-const getParentLayer = (pages: Sketch.AnyLayer, lastIndexPath: IndexPath) => {
-  return Layers.access(pages, lastIndexPath.slice(0, -1)) as Sketch.Page;
-};
-const addToParentLayer = (
-  pages: Sketch.AnyLayer,
-  selectedIndexPaths: IndexPath[],
-  element: Exclude<Sketch.AnyLayer, { _class: 'page' }>,
-) => {
-  const lastIndexPath = selectedIndexPaths[selectedIndexPaths.length - 1];
-  const parent = getParentLayer(pages, lastIndexPath);
+const getParentLayer = (page: Sketch.AnyLayer, indexPath: IndexPath) =>
+  Layers.access(page, indexPath.slice(0, -1)) as Layers.ParentLayer;
 
-  parent.layers.splice(lastIndexPath[lastIndexPath.length - 1], 0, element);
+const addSiblingLayer = (
+  page: Sketch.AnyLayer,
+  indexPath: IndexPath,
+  layer: Layers.ChildLayer,
+) => {
+  const parent = getParentLayer(page, indexPath);
+
+  parent.layers.splice(indexPath[indexPath.length - 1], 0, layer);
 };
 
-const getReversedDirectChildrensLayersIndexPaths = (
-  state: ApplicationState,
-  ids: string[],
-) => {
-  return (
-    getLayerIndexPathsExcludingDescendants(state, ids)
-      .filter((indexPath) =>
-        Layers.isChildLayer(Layers.access(getCurrentPage(state), indexPath)),
-      )
-      // Reverse the indexPaths to simplify deletion
-      .reverse()
+const getIndexPathsForGroup = (state: ApplicationState, ids: string[]) => {
+  return getLayerIndexPathsExcludingDescendants(
+    state,
+    ids,
+  ).filter((indexPath) =>
+    Layers.isChildLayer(Layers.access(getCurrentPage(state), indexPath)),
   );
 };
 
-const createGroup = (
-  action: 'group' | 'symbolMaster',
+const createGroup = <T extends Sketch.Group | Sketch.SymbolMaster>(
+  model: T,
   page: Sketch.Page,
   ids: string[],
   name: string,
-  selectedIndexPaths: IndexPath[],
-): Sketch.Group | Sketch.SymbolMaster | undefined => {
-  const lastIndexPath = selectedIndexPaths[selectedIndexPaths.length - 1];
+  indexPaths: IndexPath[],
+): T | undefined => {
+  const lastIndexPath = indexPaths[0];
   const boundingRect = getBoundingRect(page, AffineTransform.identity, ids, {
     clickThroughGroups: true,
     includeHiddenLayers: true,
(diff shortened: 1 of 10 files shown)


packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -195,6 +182,136 @@ export function layerReducer(
...
+        draft.do_objectID = uuid();
+        draft.name = name;
+        draft.frame = symbolMasters.frame;
+        draft.style = produce(Models.style, (s) => {
+          s.do_objectID = uuid();
+        });
+        draft.symbolID = symbolMasters.symbolID;
+      });
+
+      const createSymbolsPage = (
```
dabbott, line 274: This could be a helper function that we export from `pageReducer.ts`. E.g. `export function createPage(...)`. And we could also call that within `addPage`

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -195,6 +182,136 @@ export function layerReducer(
         updateSelection(draft.selectedObjects, id, selectionType);
       });
     }
+    case 'addSymbol': {
+      const [, id, name] = action;
```
dabbott, line 186: We should definitely figure out a way to share a lot of this logic with creating groups. Maybe we can start with a helper function in this file?

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -195,6 +182,136 @@ export function layerReducer(
...
+          return page;
+        });
+
+        user[newPage.do_objectID] = {
+          scrollOrigin: '{0, 0}',
+          zoomValue: 1,
+        };
+
+        pages.push(newPage);
+        return pages[pages.length - 2];
```
dabbott, line 290: Should this be `pages[pages.length - 1]`? It looks like this is returning the page _before_ the symbols page

dabbott (changes requested):
Looks good! I did a quick test, and it worked correctly, except that the symbol master was inserted in the page _before_ the Symbols page - but I left a comment on what I think is causing that

packages/noya-state/src/reducers/layerReducer.ts
milg15: oh yess! I was doing something with it did notice I left it

packages/noya-state/src/reducers/layerReducer.ts
milg15: I'll make a couple of smaller functions of the shared actions and then we can see which ones need to be together

milg15:
Okey I made a couple of functions 😅 Some might not even be needed but I just wanted to check with you which one do you think would be more useful/necessary.

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -2,28 +2,129 @@ import Sketch from '@sketch-hq/sketch-file-format-ts';
...
+      .filter((indexPath) =>
+        Layers.isChildLayer(Layers.access(getCurrentPage(state), indexPath)),
+      )
+      // Reverse the indexPaths to simplify deletion
+      .reverse()
+  );
+};
+
+const createGroup = (
+  action: 'group' | 'symbolMaster',
```
dabbott, line 61: Let's pass in the `Models.symbolMaster` and `Models.group` instead of this action-based approach. That way we can use a generic type:

```ts
const createGroup = <T extends Sketch.Group | Sketch.SymbolMaster>(
  layer: T,
  page: Sketch.Page,
  ids: string[],
  name: string,
  indexPaths: IndexPath[],
): T | undefined => { ... }
```

We can move the `symbolID` assignment outside of this function

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -2,28 +2,129 @@ import Sketch from '@sketch-hq/sketch-file-format-ts';
...
   | [type: 'groupLayer', layerId: string | string[], name: string]
   | [type: 'ungroupLayer', layerId: string | string[]]
+  | [type: 'addSymbol', layerId: string | string[], name: string]
   | [
       type: 'selectLayer',
       layerId: string | string[] | undefined,
       selectionType?: SelectionType,
     ];
 
+const getParentLayer = (pages: Sketch.AnyLayer, lastIndexPath: IndexPath) => {
```
dabbott, line 32: I like this function, but I think we can improve the parameter names and types:

```ts
const getParentLayer = (page: Sketch.Page, indexPath: IndexPath) => {
  return Layers.access(page, indexPath.slice(0, -1)) as Layers.ParentLayer;
};
```

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -2,28 +2,129 @@ import Sketch from '@sketch-hq/sketch-file-format-ts';
...
+  return Layers.access(pages, lastIndexPath.slice(0, -1)) as Sketch.Page;
+};
+const addToParentLayer = (
+  pages: Sketch.AnyLayer,
+  selectedIndexPaths: IndexPath[],
+  element: Exclude<Sketch.AnyLayer, { _class: 'page' }>,
```
dabbott, line 38: Howabout:

```ts
const addSiblingLayer = (
  page: Sketch.Page,
  indexPath: IndexPath,
  layer: Layers.ChildLayer,
) => {
  const parent = getParentLayer(page, indexPath);

  parent.layers.splice(indexPath[indexPath.length - 1], 0, layer);
};
```

I changed the param names, and then I think this function is clearer with only a single `IndexPath`. So we can choose the last index path outside of this function

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -2,28 +2,129 @@ import Sketch from '@sketch-hq/sketch-file-format-ts';
...
+) => {
+  const lastIndexPath = selectedIndexPaths[selectedIndexPaths.length - 1];
+  const parent = getParentLayer(pages, lastIndexPath);
+
+  parent.layers.splice(lastIndexPath[lastIndexPath.length - 1], 0, element);
+};
+
+const getReversedDirectChildrensLayersIndexPaths = (
```
dabbott, line 46: Maybe `const getIndexPathsForGroup`?

packages/noya-state/src/selectors/layerSelectors.ts
```diff
@@ -64,3 +65,14 @@ export const makeGetPageLayers = (
       .map((id) => page.layers.find((layer) => layer.do_objectID === id))
       .filter((layer): layer is PageLayer => !!layer);
 };
+
+export const deleteLayers = (layers: IndexPath[], page: Sketch.Page) => {
+  layers.forEach((indexPath) => {
```
dabbott, line 70: Now that we have a separate `deleteLayers` function, maybe we can move the `.reverse` into here? Reversing is only needed for deletion anyway, so it'd be simpler if other functions didn't need to consider it.

Also, we can reverse the parameter order and rename to match other functions.

```ts
export const deleteLayers = (page: Sketch.Page, indexPaths: IndexPath[]) => {
  const reversed = [...indexPaths].reverse()
  reversed.forEach(...)
}
```

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -195,6 +220,56 @@ export function layerReducer(
...
+        ids,
+      );
+
+      const symbolMasters = createGroup(
+        'symbolMaster',
+        page,
+        ids,
+        name,
+        selectedIndexPaths,
+      ) as Sketch.SymbolMaster;
```
dabbott, line 242: If we make `createGroup` generic, we can remove the `as` cast here

dabbott (changes requested):
Sharing code via functions makes this a lot simpler! I left some comments about how I think we can improve those functions

packages/noya-state/src/reducers/layerReducer.ts
milg15: sounds good!

packages/noya-state/src/reducers/layerReducer.ts
milg15: Pages can have any except for pages 🤔  'artboard' | 'symbolMaster'

packages/noya-state/src/selectors/layerSelectors.ts
milg15: I hope a re-order it correctly, because I was using the last value for the list in some part because it was reversed now It's the first value 🤔

packages/noya-state/src/reducers/layerReducer.ts
dabbott: Oh yeah, that's true. Ok, your type was better, let's leave it with that

packages/noya-state/src/selectors/layerSelectors.ts
dabbott: Oh yeah, that's true. But hopefully it will be simpler without reversing in the first place

milg15:
I think I can place getParentLayer, and addSiblingLayer into layerSelectors. getIndexPathsForGroup to indexPathSelectors. But I am not sure on where to move createGroup

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -42,17 +129,8 @@ export function layerReducer(
...
       return produce(state, (draft) => {
-        reversed.forEach((indexPath) => {
-          const childIndex = indexPath[indexPath.length - 1];
-          const parent = Layers.access(
-            draft.sketch.pages[pageIndex],
-            indexPath.slice(0, -1),
-          ) as Layers.ParentLayer;
-          parent.layers.splice(childIndex, 1);
-        });
+        deleteLayers(indexPaths, draft.sketch.pages[pageIndex]);
```
dabbott, line 133: There's a comment above this that should go into `deleteLayer`

```ts
// We delete in reverse so that the indexPaths remain accurate even
// after some layers are deleted.
```

dabbott (approved):
That sounds good! I agree that `createGroup` can stay here, cause I don't see that being used elsewhere.

Now that our reducers are split into different files, I think it's fine if we start to have a few things functions in the same file. We could even leave the others here until we know they're going to be used elsewhere... but I think I prefer moving them now

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -2,28 +2,115 @@ import Sketch from '@sketch-hq/sketch-file-format-ts';
...
+  name: string,
+  indexPaths: IndexPath[],
+): T | undefined => {
+  const boundingRect = getBoundingRect(page, AffineTransform.identity, ids, {
+    clickThroughGroups: true,
+    includeHiddenLayers: true,
+  });
+
+  if (!boundingRect) {
+    console.info('[groupLayer] Selected layers not found');
```
dabbott, line 67: Since we log here, we can probably remove the `console.info` call in both places that call `createGroup`

milg15:
I'll merge it then, the last commit was only to move the functions

Issue #54, opened 2021-05-08
Quick measure


dabbott:
## Feature

When the user holds the <kbd>alt</kbd> key, we should display helpful measurement "guides" indicating the distance between layers. This feature is a key part of the "smart snap" feature.

There are several branches of logic to consider:
- If there are layer(s) selected:
  - If the user's mouse is *not* hovering over another layer, then:
    - If the layer is within an artboard, show the distance to the artboard's frame (unless the edge of the layer is outside of the artboard's frame):
      ![Screenshot_5_7_21__8_19_PM](https://user-images.githubusercontent.com/1198882/117524220-78ee9900-af71-11eb-9bb3-3d1877006166.png)
      ![Screenshot_5_7_21__8_37_PM](https://user-images.githubusercontent.com/1198882/117524816-1a76ea00-af74-11eb-8e61-32565a2bda9a.png)
    - ~~If the layer is _not_ within an artboard, show no guides~~ We should still show guides on layers outside artboards, even though they don't snap
  - If the user hovers their mouse over another layer, we should display the distance to each axis of the hovered layer. We should add guides extrapolating out the edges of the hovered layer if needed:
    ![Screenshot_5_7_21__8_23_PM](https://user-images.githubusercontent.com/1198882/117524453-1ea20800-af72-11eb-9b78-fe42e4e891c7.png)
    ![Screenshot_5_7_21__8_25_PM](https://user-images.githubusercontent.com/1198882/117524487-4ee9a680-af72-11eb-81a5-0688b3e3cfe7.png)
    ![Screenshot_5_7_21__8_27_PM](https://user-images.githubusercontent.com/1198882/117524562-c15a8680-af72-11eb-9574-208fe62ba6cb.png)
    ![Screenshot_5_7_21__8_29_PM](https://user-images.githubusercontent.com/1198882/117524601-ed760780-af72-11eb-9997-db9197282241.png)
- If there are no layers selected:
  - If the user's mouse is *not* hovering over another layer, show no guides
  - If the hovers their mouse over a layer, *and* that layer is within an artboard, show the distance to the artboard
    ![Screenshot_5_7_21__8_40_PM](https://user-images.githubusercontent.com/1198882/117524881-7f324480-af74-11eb-91c6-99dc4398ce75.png)

## Implementation notes:

- We can handle the X and Y axis independently - they don't influence each other at all
[...]

PR #67, opened 2021-05-14
Quick measure

leannepepper:
Implements initial quick measure feature.
Screen Shot 2021-05-14 at 1 16 51 PM

with the exception of layers in an artboard measuring to the edge of the artboard by default.

closes issue #54


packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -193,16 +218,38 @@ export default memo(function SketchFileRenderer() {
...
-          layer={layer}
-          paint={highlightPaint}
-        />
+        <>
+          <HoverOutline
+            transform={layerTransform}
+            layer={layer}
+            paint={highlightPaint}
+          />
+          {selectedAndHighlightedLayers &&
```
dabbott, line 234: Maybe it would be clearer if we rendered these elements separately from the `HoverOutline` in another `useMemo`?

In that case, we could potentially avoid a separate `useCallback` for `getSelectedAndHighlightedLayerBoundingRec`, and get the bounding rect / selected rect right before rendering these elements.

There might be some logic that we can extract out and share between both `useMemo`s, e.g. finding the layer transform. There might also be some logic that only applies to one or the other: e.g. the `highlightedLayer.precedence` only applies to the `HoverOutline`, but not the distance labels - so we shouldn't do an early return based on `highlightedLayer.precedence` when rendering `DistanceLabelAndPath`

packages/noya-renderer/src/components/DistanceLabelAndPath.tsx
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  const { CanvasKit } = useReactCanvasKit();
+  const paint = new CanvasKit.Paint();
+  paint.setColor(CanvasKit.Color4f(0.52, 0.248, 1.0));
+  paint.setPathEffect(CanvasKit.PathEffect.MakeDash([1, 2]));
+  paint.setStyle(CanvasKit.PaintStyle.Stroke);
+  paint.setStrokeWidth(1);
+
+  const explicitPaint = new CanvasKit.Paint();
```
dabbott, line 282: Can we use `usePaint` here?

packages/noya-renderer/src/components/DistanceLabelAndPath.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+  return (
+    <>
+      <RCKRect rect={bgRect} paint={backgroundFill} />
+      <Text rect={labelRect} paragraph={paragraph} />
+    </>
+  );
+}
+
+function MeasureAndExtensionPaths({
```
dabbott, line 74: We should definitely try to come up with an abstraction here, but I might have to fiddle with the types a bit. I'll see what I can come up with later today

packages/noya-renderer/src/components/DistanceLabelAndPath.tsx
```diff
@@ -0,0 +1,323 @@
...
+  selectedLayer,
+  highlightedLayer,
+}: {
+  selectedLayer: Rect;
+  highlightedLayer: Rect;
+}) {
+  const highlightedBounds = createBounds(highlightedLayer);
+  const selectedBounds = createBounds(selectedLayer);
+
+  let implicitLines: Point[][] = [];
```
dabbott, line 84: Do we still want to use the naming we talked about when pairing? (I'm fine with either way)

> "Extension guides" - the lines extending edges of the highlighted layer
"Measurement guides" - the lines indicating a distance between two rects, that often show a text bubble with the distance value

packages/noya-renderer/src/components/DistanceLabelAndPath.tsx
```diff
@@ -0,0 +1,323 @@
...
+  let explicitLines: Point[][] = [];
+  let distanceMeasurements: DistanceMeasurementProps[] = [];
```
dabbott, line 86: I found that sometimes the text labels appear on the opposite side of the explicit lines, when the explicit lines are drawn inward. If there's a simple fix, we should do it as part of this issue, otherwise we can come back to it.

Screen Shot 2021-05-14 at 1 55 35 PM

dabbott (changes requested):
Looks great! I tested it out and overall I think it works extremely well. 

The main thing I'd like to improve before merging is the abstraction around the 4 different edges. But that's probably going to involve some TypeScript weirdness, so I'll take a look a little later.

packages/noya-renderer/src/components/DistanceLabelAndPath.tsx
leannepepper: yes, renamed to` measureGuidePaint` and used `usePaint` hook

packages/noya-renderer/src/components/DistanceLabelAndPath.tsx
leannepepper: Oops! old habits die hard I guess, all names have been changed to `measureGuideLines` and `extensionGuideLines`

packages/noya-renderer/src/components/DistanceLabelAndPath.tsx
leannepepper: Midpoint calculation was incorrect, issue has been fixed. 
Screen Shot 2021-05-14 at 5 07 27 PM

packages/noya-renderer/src/components/SketchFileRenderer.tsx
leannepepper: Removed element from `hoverOutline` and into its own `useMemo` callback `distanceLabelAndPathBetweenSketchLayers`. However I'm less familiar with this file, so please let me know if I can improve this change. Commit: https://github.com/noya-app/noya/pull/67/commits/83b58ff164b6619d64ef380f46ace1e60e73e4df

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -195,11 +262,13 @@ export default memo(function SketchFileRenderer() {
 
     return (
       highlightedLayer && (
-        <HoverOutline
-          transform={layerTransform}
-          layer={layer}
-          paint={highlightPaint}
-        />
+        <>
```
dabbott, line 265: We can remove this fragment

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -169,6 +170,72 @@ export default memo(function SketchFileRenderer() {
...
+
+    if (!indexPath) return;
+
+    let selectedAndHighlightedLayers;
+    if (highlightedLayer.isMeasured) {
+      selectedAndHighlightedLayers = getSelectedAndHighlightedLayerBoundingRec();
+    }
+
+    return (
+      highlightedLayer && (
```
dabbott, line 218: We already check for `!highlightedLayer` above, so we can remove this check. We also shouldn't need a `<>` here. I think if we combine some of the checks into the condition above, we shouldn't need any `&&` in this return

leannepepper: Because `getBoundingRect` returns possibly an undefined value I can't find a way around adding one check in the return for `selectedAndHighlightedLayers.highlightedBoundingRec`. The other check have been removed.

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -169,6 +170,72 @@ export default memo(function SketchFileRenderer() {
...
+    );
+    return {
+      selectedBoundingRec: boundingRect,
+      highlightedBoundingRec: highlightedBoundingRec,
+    };
+  }, [boundingRect, highlightedLayer, page, state.selectedObjects.length]);
+
+  const distanceLabelAndPathBetweenSketchLayers = useMemo(() => {
+    if (
+      !highlightedLayer ||
```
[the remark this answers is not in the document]
leannepepper: consolidated functions

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -169,6 +170,49 @@ export default memo(function SketchFileRenderer() {
...
+      page,
+      AffineTransform.identity,
+      [highlightedLayer.id],
+      {
+        clickThroughGroups: true,
+        includeHiddenLayers: true,
+      },
+    );
+
+    const selectedAndHighlightedLayers = {
```
dabbott, line 201: Can we remove this object and instead pass the props directly?

Very minor: I would probably name variables with the suffix `Rect` rather than `Rec`, since the type is `Rect`. And the props `selectedLayer` and `highlightedLayer` could probably also be `selectedRect` and `highlightedRect`, for consistency (`Layer` and `Rect` are different types)

dabbott (approved):
This looks great 🔥. I think it ended up being pretty concise. I added one more comment, but I'll go ahead and approve this now

packages/noya-renderer/src/components/SketchFileRenderer.tsx
leannepepper: changed naming and removed object to pass props directly

PR #72, opened 2021-05-18
Symbol in Theme tabs

milg15:
Show grid of symbols and allow to change name but no preview

### DIFF
--- a/packages/app/src/containers/ThemeSymbolsInspector.tsx
+++ b/packages/app/src/containers/ThemeSymbolsInspector.tsx
@@ -1,0 +1,44 @@
+import { Divider } from 'noya-designsystem';
+import { Selectors } from 'noya-state';
+import { memo, useCallback } from 'react';
+import {
+  useApplicationState,
+  useSelector,
+} from '../contexts/ApplicationStateContext';
+import useShallowArray from '../hooks/useShallowArray';
+import withSeparatorElements from '../utils/withSeparatorElements';
+import NameInspector from '../components/inspector/NameInspector';
+import OpacityInspector from './OpacityInspector';
+import ShadowInspector from './ShadowInspector';
+import { delimitedPath } from 'noya-utils';
+
+export default memo(function ThemeSymbolsInspector() {
+  const [, dispatch] = useApplicationState();
+
+  const selectedSymbols = useShallowArray(
+    useSelector(Selectors.getSelectedSymbols),
+  );
+
+  const handleNameChange = useCallback(
+    (value: string) =>
+      dispatch(
+        'setSymbolName',
+        selectedSymbols.map((v) => v.do_objectID),
+        value,
+      ),
+    [dispatch, selectedSymbols],
+  );
+
+  if (selectedSymbols.length === 0) return null;
+
+  const elements = [
+    <NameInspector
+      names={selectedSymbols.map((v) => delimitedPath.basename(v.name))}
+      onNameChange={handleNameChange}
+    />,
+    <OpacityInspector />,
+    <ShadowInspector />,
+  ];
+
+  return <>{withSeparatorElements(elements, <Divider />)}</>;
+});
--- a/packages/app/src/containers/ThemeSymbolsInspector.tsx
+++ b/packages/app/src/containers/ThemeSymbolsInspector.tsx
@@ -8,8 +8,6 @@ import {
 import useShallowArray from '../hooks/useShallowArray';
 import withSeparatorElements from '../utils/withSeparatorElements';
 import NameInspector from '../components/inspector/NameInspector';
-import OpacityInspector from './OpacityInspector';
-import ShadowInspector from './ShadowInspector';
 import { delimitedPath } from 'noya-utils';
 
 export default memo(function ThemeSymbolsInspector() {
@@ -36,8 +34,6 @@ export default memo(function ThemeSymbolsInspector() {
       names={selectedSymbols.map((v) => delimitedPath.basename(v.name))}
       onNameChange={handleNameChange}
     />,
-    <OpacityInspector />,
-    <ShadowInspector />,
   ];
 
   return <>{withSeparatorElements(elements, <Divider />)}</>;
--- a/packages/noya-state/src/reducers/layerReducer.ts
+++ b/packages/noya-state/src/reducers/layerReducer.ts
@@ -17,6 +17,7 @@ import {
   addSiblingLayer,
   getSymbols,
   getSymbolsInstaces,
+  getCurrentSymbolPageIndex,
 } from '../selectors/selectors';
 import { SelectionType, updateSelection } from '../utils/selection';
 import { ApplicationState } from './applicationReducer';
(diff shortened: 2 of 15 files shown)


packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -385,6 +394,33 @@ export function themeReducer(
...
     }
+    case 'setSymbolName': {
+      const [, id, name] = action;
+
+      const ids = typeof id === 'string' ? [id] : id;
+
+      return produce(state, (draft) => {
+        const symbols: Sketch.SymbolMaster[] = [];
+
+        draft.sketch.pages.forEach((p) =>
```
dabbott, line 405: Can we call the `getSymbols` selector here? E.g. `getSymbols(state).filter(symbol => ids.includes(symbol.do_objectID)`

packages/noya-state/src/selectors/themeSelectors.ts
```diff
@@ -62,3 +62,25 @@ export function groupThemeComponents<
 
   return array;
 }
+
+export const getSymbols = (state: ApplicationState): Sketch.SymbolMaster[] => {
+  const symbols: Sketch.SymbolMaster[] = [];
+
+  state.sketch.pages.forEach((p) =>
```
dabbott, line 69: We should be able to use `flatMap` here instead of `forEach` and `push`, e.g.

```ts
const symbols = draft.sketch.pages.flatMap(p.layers.flatMap(...))
```

We can also use `Layers.isSymbolMaster(l)` instead of `._class === 'symbolMaster`

packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -29,6 +29,11 @@ export type ThemeAction =
       id: string | string[] | undefined,
       selectionType?: SelectionType,
     ]
+  | [
+      type: `selectSymbol`,
```
dabbott, line 33: Above we have the type `ComponentsElements`. Can we add `Symbol` to that? I think symbols should support all the same actions as the other theme components

But we could also do that in a separate PR if you prefer

packages/app/src/containers/ThemeSymbolsInspector.tsx
```diff
@@ -0,0 +1,44 @@
...
+  );
+
+  if (selectedSymbols.length === 0) return null;
+
+  const elements = [
+    <NameInspector
+      names={selectedSymbols.map((v) => delimitedPath.basename(v.name))}
+      onNameChange={handleNameChange}
+    />,
+    <OpacityInspector />,
```
dabbott, line 39: The SymbolMaster inspector shouldn't show opacity or shadows:
 
Pasted_Image_5_18_21__9_24_AM

The SymbolInstance inspector will though, when we get to that!

dabbott (changes requested):
This looks good! There are a few more actions that we should support for symbols, such as grouping and deleting, to be consistent with the other kinds of theme components - but we could also do that separately, I'm fine with either way

packages/noya-state/src/reducers/themeReducer.ts
milg15: I will add it, I just need to add the other actions :D

packages/noya-state/src/reducers/themeReducer.ts
milg15: does it work? don't we need tou use the draft that has the type `WritableDraft<ApplicationState>` and it can't be use in that function 🤔

packages/noya-state/src/reducers/themeReducer.ts
milg15: Oh but since addSymbol it's being handled somewhere else I don't think I can add it here anymore 🤔

packages/noya-state/src/reducers/themeReducer.ts
dabbott: I wasn't sure, so I tried it and it seems OK:

Screen Shot 2021-05-18 at 1 13 59 PM

I'm actually not sure when the `WritableDraft` stuff is necessary. I think it's a good idea to use that type on any functions that _mutate_ the state, but shouldn't matter for functions that don't.

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -91,6 +98,64 @@ const createGroup = <T extends Sketch.Group | Sketch.SymbolMaster>(
...
+        l.do_objectID = uuid();
+      }),
+    );
+  });
+
+  deleteLayers([indexPath], page);
+  addSiblingLayer(page, indexPath, group);
+};
+
+const detacthSymbolIntances = (
```
dabbott, line 143: Spelling: `detacth` => `detach`

packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -361,28 +375,52 @@ export function themeReducer(
...
-        const array =
-          action[0] === 'setSwatchName'
-            ? draft.sketch.document.sharedSwatches?.objects ?? []
-            : action[0] === 'setTextStyleName'
-            ? draft.sketch.document.layerTextStyles?.objects ?? []
-            : draft.sketch.document.layerStyles?.objects ?? [];
+        setComponentName(
+          ids,
+          name,
+          draft.sketch.document.layerTextStyles?.objects ?? [],
```
dabbott, line 387: Should this one be `.layerStyles` instead of `.layerTextStyles`?

packages/noya-state/src/selectors/themeSelectors.ts
```diff
@@ -62,3 +73,62 @@ export function groupThemeComponents<
...
+
+  const filter =
+    getCurrentTab(state) === 'canvas'
+      ? state.selectedObjects
+      : state.selectedSymbolsIds;
+
+  return symbols.filter((symbol) => filter.includes(symbol.do_objectID));
+};
+
+export const getSymbolsInstacesIndexPaths = (
```
dabbott, line 113: Spelling: `Instaces` => `Instances`. Same with the function below

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -107,9 +172,18 @@ export function layerReducer(
         ids.includes(layer.do_objectID),
       );
 
-      // We delete in reverse so that the indexPaths remain accurate even
-      // after some layers are deleted.
+      const symbolsInstancesIndexPaths = getSelectedSymbols(
```
dabbott, line 175: This `deleteLayer` action looks very similar to `deleteSymbol`. Can we remove `deleteSymbol` entirely, and just dispatch this action instead?

dabbott (changes requested):
Wow, this is great! All the helper functions make this much easier to follow

packages/noya-state/src/selectors/themeSelectors.ts
```diff
@@ -62,3 +73,62 @@ export function groupThemeComponents<
...
+  array: T[],
+) {
+  array.forEach((object: ComponentsTypes) => {
+    if (!ids.includes(object.do_objectID)) return;
+
+    const group = delimitedPath.dirname(object.name);
+    object.name = delimitedPath.join([group, name]);
+  });
+
+  return array;
```
dabbott, line 89: I would probably remove the return here, since it's uncommon to return the same object we pass as an argument

packages/noya-state/src/reducers/themeReducer.ts
milg15: good catch! I must have missed it

packages/noya-state/src/reducers/layerReducer.ts
milg15: The main difference is that in the deleteSymbols goes trough every page and the deleteLayer only goes trough the main page. How should I go about changing this part? Do I make it general

dabbott: Ah right. I think it's OK to go through every page in `deleteLayer`. So we can move the `deleteSymbols` implementation into `deleteLayer`, and remove `deleteSymbols`.

I want to move away from relying on the current page in actions in general, so I like that approach better anyway. A lot of the actions that use the current page were some of the first ones I wrote, but it's less flexible than searching every page

dabbott (approved):
Looks great! 🔥

PR #75, opened 2021-05-20
Render symbols in themes tab. LayerList improvements

dabbott:
Screen Shot 2021-05-20 at 10 16 14 AM

Screen Shot 2021-05-20 at 10 16 47 AM

@milg15


milg15 (approved):
Looks good the only thing I might look into the future it's finding a way of showing the letter in a sharper way. The texts looks a bit blurry 🤔 but I don't really know what would be a solution for that.

PR #78, opened 2021-05-20
Duplicate symbols, pages and layers

milg15:


packages/noya-state/src/reducers/pageReducer.ts
```diff
@@ -80,7 +85,9 @@ export function pageReducer(
 
           Layers.visit(page, (layer) => {
             layer.do_objectID = uuid();
+
             if (layer.style) layer.style.do_objectID = uuid();
+            if (layer._class === 'symbolMaster') layer.symbolID = uuid();
```
dabbott, line 90: We also want to update any instances within this page that reference the symbolMaster. I might do this in 2 passes: first we figure out all the new symbolIDs, then we assign them.

```ts
const mappedSymbolIDs: Record<string, string> = {}

page.layers.filter(Layers.isSymbolMaster).forEach(symbolMaster => {
  mappedSymbolIDs[symbolMaster.symbolID] = uuid()
})

Layers.visit(page, (layer) => {
  // ...
  if (layer._class === 'symbolMaster' 
      || (layer._class === 'symbolInstance' && layer.symbolID in mappedSymbolIDs)) { 
    layer.symbolID = mappedSymbolIDs[layer.symbolID]
  }
})
```

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -288,6 +289,48 @@ export function layerReducer(
...
+            const elem = produce(layer, (layer) => {
+              layer.name = layer.name + ' Copy';
+
+              Layers.visit(layer, (layer) => {
+                if (layer.style) layer.style.do_objectID = uuid();
+                if (layer._class === 'symbolMaster') {
+                  layer.symbolID = uuid();
+                  layer.frame = {
+                    ...layer.frame,
+                    ...{
```
dabbott, line 319: We can remove the extra object `...{` here

packages/app/src/containers/LayerList.tsx
```diff
@@ -207,9 +207,34 @@ export default memo(function LayerList() {
...
+  const canBeSymbol = useCallback((): number => {
+    const selectedItems = items.filter((i) => selectedObjects.includes(i.id));
+
+    if (
+      selectedItems.length === 0 ||
+      selectedItems.some((l) => l.type === 'symbolMaster')
+    )
+      return 0;
+
+    if (
```
dabbott, line 219: It looks like the same condition for returning `1` and `2`? Are these supposed to be different? If not, we should return a boolean

milg15: yes, they are supposed to be different, this could be handle by a type but I am not sure yet. It's because when you have artboard you don't ask for the symbol name 🤔

dabbott: We can probably make a separate boolean variable, like `shouldAskForSymbolName`, rather than re-using `canBeSymbol`. Ideally `canBeSymbol` will also be a boolean variable, rather than a function

packages/noya-state/src/selectors/indexPathSelectors.ts
```diff
@@ -74,3 +74,13 @@ export const getIndexPathsForGroup = (
     Layers.isChildLayer(Layers.access(getCurrentPage(state), indexPath)),
   );
 };
+
+export const getIndexofArboardLayers = (
```
dabbott, line 78: `getIndexofArboardLayers` => `getIndexPathsOfArtboardLayers`

packages/noya-state/src/selectors/layerSelectors.ts
```diff
@@ -95,3 +96,21 @@ export const addSiblingLayer = <
...
   parent.layers.splice(indexPath[indexPath.length - 1], 0, ...l);
 };
+
+export const getLeftMostLayerPoint = (page: Sketch.Page) => {
+  const point = {
+    x: 0,
+    y: 0,
+  };
+
+  if (!page) return point;
```
dabbott, line 106: `page` isn't optional, so we shouldn't need this check

packages/noya-state/src/selectors/layerSelectors.ts
```diff
@@ -95,3 +96,21 @@ export const addSiblingLayer = <
   const l = layer instanceof Array ? layer : [layer];
   parent.layers.splice(indexPath[indexPath.length - 1], 0, ...l);
 };
+
+export const getLeftMostLayerPoint = (page: Sketch.Page) => {
```
dabbott, line 100: For this function, I'd recommend sorting by `layer.frame.x` and choosing the first layer in the sorted array. I think the logic will be simpler

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -261,33 +286,81 @@ export function layerReducer(
...
-        draft.do_objectID = uuid();
-        draft.name = name;
-        draft.frame = symbolMasters.frame;
-        draft.style = produce(Models.style, (s) => {
-          s.do_objectID = uuid();
-        });
-        draft.symbolID = symbolMasters.symbolID;
-      });
+
+      const prevFrame = { ...symbolMasters.frame };
```
dabbott, line 291: Maybe `originalFrame`?

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -242,16 +247,36 @@ export function layerReducer(
...
+          deleteLayers(indexPathsArtboards, pages[pageIndex]);
+          indexPathsArtboards.forEach((indexPath: IndexPath) => {
+            const artboard = Layers.access(page, indexPath);
+            const symbolMaster = {
+              ...Models.symbolMaster,
+              ...artboard,
+            } as Sketch.SymbolMaster;
+
+            symbolMaster._class = 'symbolMaster';
+            symbolMaster.symbolID = uuid();
```
dabbott, line 272: Maybe

```ts
const artboard = Layers.access(page, indexPath) as Sketch.Artboard;

const symbolMaster = {
  ...Models.symbolMaster,
  ...artboard,
  _class: 'symbolMaster' as const,
  symbolID: uuid()
};
```

dabbott (changes requested):
I tested briefly and I think everything works well! There are a few small improvements in the code we can try

packages/noya-state/src/reducers/layerReducer.ts
milg15: oh the as const ! I was getting an error before

packages/app/src/containers/LayerList.tsx
```diff
@@ -207,9 +207,27 @@ export default memo(function LayerList() {
...
     items.find((i) => i.id === selectedObjects[0])?.type === 'symbolInstance';
 
+  const canBeSymbol = useMemo(() => {
+    const selectedItems = items.filter((i) => selectedObjects.includes(i.id));
+
+    return (
+      selectedItems.length >= 1 &&
+      !selectedItems.some((l) => l.type === 'symbolMaster') &&
+      ((selectedItems[0].type === 'artboard' &&
+        selectedItems.every((l) => l.type === 'artboard')) ||
```
dabbott, line 217: If we're already checking that `every` item is an artboard, can we skip checking the first item, `selectedItems[0].type === 'artboard'`?

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -261,33 +286,84 @@ export function layerReducer(
...
+          const draftPage = draft.sketch.pages[pageIndex];
+
+          indexPaths.forEach((indexPath) => {
+            const layer = Layers.access(page, indexPath);
+            const elem = produce(layer, (layer) => {
+              layer.name = layer.name + ' Copy';
+
+              Layers.visit(layer, (layer) => {
+                if (layer.style) layer.style.do_objectID = uuid();
+                if (layer._class === 'symbolMaster') {
```
dabbott, line 350: Not super important but I usually go with `Layers.isSymbolMaster` instead of checking `_class`

packages/noya-state/src/reducers/pageReducer.ts
```diff
@@ -75,12 +80,26 @@ export function pageReducer(
...
+
         const duplicatePage = produce(page, (page) => {
           page.name = `${page.name} Copy`;
 
           Layers.visit(page, (layer) => {
             layer.do_objectID = uuid();
             if (layer.style) layer.style.do_objectID = uuid();
+
+            if (
+              layer._class === 'symbolMaster' ||
```
dabbott, line 97: Same as above, slight preference for `Layers.isSymbolMaster` and `Layers.isSymbolInstance` over `_class` check

packages/noya-state/src/reducers/pageReducer.ts
```diff
@@ -99,13 +118,23 @@ export function pageReducer(
       const page = getCurrentPage(state);
       const pageIndex = getCurrentPageIndex(state);
 
+      const symbolsIds = page.layers.flatMap((layer) =>
+        layer._class === 'symbolMaster' ? [layer.symbolID] : [],
```
dabbott, line 122: `Layers.isSymbolMaster`

packages/noya-state/src/selectors/indexPathSelectors.ts
```diff
@@ -74,3 +74,13 @@ export const getIndexPathsForGroup = (
...
   );
 };
+
+export const getIndexPathsOfArtboardLayers = (
+  state: ApplicationState,
+  ids: string[],
+) => {
+  return getLayerIndexPathsExcludingDescendants(state, ids).filter(
+    (indexPath) =>
+      Layers.access(getCurrentPage(state), indexPath)._class === 'artboard',
```
dabbott, line 84: And here we could create a `Layers.isArtboard`

packages/noya-state/src/selectors/layerSelectors.ts
```diff
@@ -95,3 +96,14 @@ export const addSiblingLayer = <
...
```
dabbott, line 100: Should this be called `getRightMostLayerBounds`? 

- Using `maxX` would give us the right edge. `minX` or just `frame.x` would give us left
- We're returning a Bounds type, instead of a Point

dabbott (approved):

packages/noya-state/src/reducers/layerReducer.ts
milg15: we are using LayerListItem not Layers it's not working 🤔

dabbott: Do you mean in`LayerList.tsx`? You're right, I don't think we can use these functions there

milg15: Oh I thought this was a different file I just added this change

PR #80, opened 2021-05-21
Render symbols with overrides

dabbott:
This renders all 5 kinds of overrides: text, style, text style, image, and symbol instance. It also includes utilities which we can potentially reuse for rendering the symbol inspectors.

Here are some example symbols:

Screen Shot 2021-05-21 at 11 06 50 AM

And with overrides:

Screen Shot 2021-05-21 at 11 06 59 AM

@milg15


milg15 (approved):
It looks really nice! I ran into some render issues but I don't think they are related to this changes.

PR #82, opened 2021-05-21
Add drawn layers into artboards

dabbott:
@milg15


milg15 (commented):
I am not sure if this is something that you should do in this PR but maybe when you move something outside of the artboard should it detach of the artboard?

milg15:
image
image
Example

milg15:
Expected behavior (I just drew a rectangle in the same place) 
image

dabbott:
This is super important! Since it'll mostly exist in the interactionReducer again, I think I'm gonna wait for Leanne's PR first though

PR #83, opened 2021-05-22
Symbols Inspectors

milg15:
Symbol Inspector for theme and canvas SymbolsMasters. 

- Only works on one SymbolMaster at a time.
- It doesn't show override properties for layers.

image


packages/app/src/components/inspector/SymbolOverrideRow.tsx
```diff
@@ -0,0 +1,30 @@
+import { Spacer } from 'noya-designsystem';
+import { memo } from 'react';
+import * as InspectorPrimitives from './InspectorPrimitives';
+
+interface Props {
+  allowsOverrides: boolean;
+  setAllowOverride: (value: boolean) => void;
```
dabbott, line 7: `setAllowOverride` => `setAllowsOverrides` for consistency

packages/app/src/components/inspector/SymbolOverrideRow.tsx
```diff
@@ -0,0 +1,30 @@
...
+}: Props) {
+  return (
+    <InspectorPrimitives.Section>
+      <InspectorPrimitives.Title>Manage Overrides</InspectorPrimitives.Title>
+      <Spacer.Vertical size={2} />
+      <InspectorPrimitives.Row>
+        <InspectorPrimitives.Checkbox
+          type="checkbox"
+          checked={allowsOverrides}
+          onChange={(evt) => setAllowOverride(evt.target.checked)}
```
dabbott, line 22: useCallback

packages/app/src/components/inspector/SymbolSourceRow.tsx
```diff
@@ -0,0 +1,92 @@
...
+interface Props {
+  resizesContent: boolean;
+  backgroundColor: Sketch.Color;
+  hasBackgroundColor: boolean;
+  includeBackgroundColorInExport: boolean;
+  includeBackgroundColorInInstance: boolean;
+  setAdjustContentOnResize: (value: boolean) => void;
+  setHasBackgroundColor: (value: boolean) => void;
+  setBackgroundColor: (value: Sketch.Color) => void;
+  setIncludeBgInExport: (value: boolean) => void;
```
dabbott, line 16: We don't abbreviate "background" as "bg" anywhere else, so for consistency, we shouldn't abbreviate it here or in our actions

packages/app/src/components/inspector/SymbolSourceRow.tsx
```diff
@@ -0,0 +1,92 @@
...
+}: Props) {
+  return (
+    <InspectorPrimitives.Section>
+      <InspectorPrimitives.Title>Symbol Source</InspectorPrimitives.Title>
+      <Spacer.Vertical size={4} />
+      <InspectorPrimitives.Row>
+        <InspectorPrimitives.Checkbox
+          type="checkbox"
+          checked={resizesContent}
+          onChange={(evt) => setAdjustContentOnResize(evt.target.checked)}
```
dabbott, line 40: `useCallback` throughout this file

packages/app/src/containers/Inspector.tsx
```diff
@@ -83,6 +84,12 @@ export default memo(function Inspector() {
 
     const hasTextLayer = selectedLayers.some((l) => Layers.isTextLayer(l));
     const hasAllTextLayer = selectedLayers.every((l) => Layers.isTextLayer(l));
+    const hasSymbolMaster = selectedLayers.some((l) =>
+      Layers.isSymbolMaster(l),
+    );
+    const hasOneSymbolMaster =
```
dabbott, line 90: `hasOneSymbolMaster = hasSymbolMaster && selectedLayers.length === 1`

packages/app/src/containers/SymbolInspector.tsx
```diff
@@ -0,0 +1,97 @@
+import { Selectors } from 'noya-state';
```
dabbott, line 1: We should rename this file to `SymbolMasterInspector`, since we're also going to have a `SymbolInstanceInspector`

packages/app/src/containers/ThemeSymbolsInspector.tsx
```diff
@@ -29,11 +30,14 @@ export default memo(function ThemeSymbolsInspector() {
 
   if (selectedSymbols.length === 0) return null;
 
+  const showSymbolsInspector = selectedSymbols.length === 1;
```
dabbott, line 33: Do we need this? Within `SymbolInspector` we already make sure we only have 1

packages/noya-state/src/reducers/symbolsReducer.ts
```diff
@@ -0,0 +1,175 @@
...
+  | [type: 'setMinWidth', value: number]
+  | [type: 'setAllowsOverride', value: boolean];
+export function symbolsReducer(
+  state: ApplicationState,
+  action: SymbolsAction,
+): ApplicationState {
+  switch (action[0]) {
+    case 'setAdjustContentOnResize': {
+      const [, value] = action;
+      const filter =
```
dabbott, line 25: I think we can use `getSelectedSymbols` throughout this file to handle the filtering and current tab check

packages/noya-state/src/reducers/symbolsReducer.ts
```diff
@@ -0,0 +1,175 @@
...
+        getCurrentTab(state) === 'canvas'
+          ? state.selectedObjects
+          : state.selectedSymbolsIds;
+
+      return produce(state, (draft) => {
+        const symbol = getSymbols(draft).filter((symbol) =>
+          filter.includes(symbol.do_objectID),
+        )[0];
+        symbol.groupLayout =
+          value === undefined || Number.isNaN(value)
```
dabbott, line 105: We shouldn't pass around `NaN` values. If we're parsing and getting a `NaN`, we should use a different approach, e.g. passing `undefined`. `NaN` can be tricky to work, so I'd treat it more like an error case and handle it as soon as it pops up

packages/noya-state/src/reducers/symbolsReducer.ts
```diff
@@ -0,0 +1,175 @@
...
+      const [, value] = action;
+      const filter =
+        getCurrentTab(state) === 'canvas'
+          ? state.selectedObjects
+          : state.selectedSymbolsIds;
+
+      return produce(state, (draft) => {
+        const symbol = getSymbols(draft).filter((symbol) =>
+          filter.includes(symbol.do_objectID),
+        )[0];
```
dabbott, line 33: Checking for `[0]` should be safe given that we should never call this without a selection, but since eventually we want to handle multiple symbols, and since it's not really more effort, I might just update all symbols in a forEach.

```ts
const symbols = ...

symbols.forEach(symbol => {
  symbol.resizesContent = value;
})
```

dabbott (changes requested):
👍 I left some feedback, but it's all pretty minor

packages/app/src/components/inspector/SymbolLayoutRow.tsx
```diff
@@ -0,0 +1,165 @@
...
+  setMinWidth: (value: number) => void;
+}
+
+export default memo(function SymbolInspector({
+  groupLayout,
+  setMinWidth,
+  setGroupLayout,
+  setLayoutAnchor,
+}: Props) {
+  const layoutValue = useMemo(() => {
```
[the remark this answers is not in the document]
milg15: I did what you show for the most part. I only left this line like this instead of placing it with the default Horizontal because it didn't allow to have the value as undefined for showing 'none'

```ts
const layoutAxis = inferredLayout?.axis;
```

I really like how it looks nice way nicer

packages/app/src/containers/ThemeSymbolsInspector.tsx
milg15: I think we should keep it for now since the inspector it's currently only working on the first symbol. Unless it should show even when selecting multiple symbols

milg15: I don't have any way of showing conflicting values, in the fields I am showing 🤔 Like when you we show "Multiples" in the input when you are selecting multiple things but they don't share the same value for that field

dabbott: Oh yeah, good point, we should keep it

packages/app/src/components/inspector/SymbolLayoutRow.tsx
```diff
@@ -0,0 +1,174 @@
...
+} from 'noya-designsystem';
+import { Layers } from 'noya-state';
+import { memo, useCallback } from 'react';
+import * as InspectorPrimitives from './InspectorPrimitives';
+
+interface Props {
+  groupLayout?: Sketch.FreeformGroupLayout | Sketch.InferredGroupLayout;
+  setLayoutAxis: (value: Sketch.InferredLayoutAxis | undefined) => void;
+  setLayoutAnchor: (value: Sketch.InferredLayoutAnchor) => void;
+  setMinWidth: (value: number, mode: 'replace' | 'adjust') => void;
```
dabbott, line 26: Can we use the `SetNumberMode` type for `mode`?

packages/app/src/components/inspector/SymbolLayoutRow.tsx
```diff
@@ -0,0 +1,174 @@
...
+          </RadioGroup.Item>
+        </RadioGroup.Root>
+      </InspectorPrimitives.Row>
+      <Spacer.Vertical size={6} />
+
+      <InspectorPrimitives.Row>
+        <LabeledElementView renderLabel={renderLabel}>
+          <RadioGroup.Root
+            id={`${layoutAxis}-${layoutAnchor}`}
+            value={layoutAnchor?.toString() ?? 'none'}
```
dabbott, line 110: I think it's good like this, but I still like the idea of using `Horizontal` as a default, since visually that's what we show even when "none" is selected for the axis. Can we do the "none" based on `isAnchorDisabled?` E.g. 

```ts
value={isAnchorDisabled ? undefined : layoutAnchor.toString()}
```

The `disabled` prop should already prevent us from calling the callback with an invalid value. I think we can also use `undefined` instead of `"none"` in this case, since we don't need to show a `"none"` RadioGroup.Item

packages/noya-state/src/reducers/symbolsReducer.ts
```diff
@@ -0,0 +1,148 @@
...
+        symbols.forEach((symbol) => {
+          symbol.groupLayout =
+            value === undefined
+              ? { _class: 'MSImmutableFreeformGroupLayout' }
+              : {
+                  _class: 'MSImmutableInferredGroupLayout',
+                  axis: value,
+                  layoutAnchor:
+                    symbol.groupLayout &&
+                    symbol.groupLayout._class ===
```
dabbott, line 90: Can we use our `isInferredLayout` here and below?

packages/noya-state/src/layers.ts
```diff
@@ -68,6 +68,12 @@ export const isSymbolInstance = (
   return layer._class === 'symbolInstance';
 };
 
+export const isInferredLayout = (
```
dabbott, line 71: I would move this into a new file `groupLayout.ts` in the same directory. Everything in this file takes `Layer` types as the first parameter. We'll probably have other layout-related functions in the future

dabbott (approved):
This is looking great! I found a few other small things, but I'll go ahead and approve this now

packages/app/src/components/inspector/SymbolLayoutRow.tsx
milg15: `isAnchorDisabled checks if layoutAxis === undefined` if I do this `const layoutAxis =
    inferredLayout?.axis ?? Sketch.InferredLayoutAxis.Horizontal;` it will never be false 🤔

milg15: using inferredLayout  instead of the layoutAxis fixes this issue  👍

milg15: oh if i use undefined instead of a string the  radio does this weird thing were it just puts a random value 🤔 and selects something even when it shouldn't
with `undefined` (`isAnchorDisabled ? undefined : layoutAnchor?.toString()`)
image
putting a string (`isAnchorDisabled ? '' : layoutAnchor?.toString()`)
image

PR #90, opened 2021-05-27
Symbol Instance Inspector

milg15:


milg15:
Ignore the `SymbolInstanceOverridesRow.tsx` file 😆

packages/noya-state/src/selectors/layerSelectors.ts
```diff
@@ -26,7 +27,9 @@ export const getSelectedTextLayers = (
   );
 };
 
-export const getSelectedLayers = (state: ApplicationState): PageLayer[] => {
+export const getSelectedLayers = (
+  state: ApplicationState | Draft<ApplicationState>,
```
dabbott, line 31: I think just `Draft<ApplicationState>` will be enough here. It seems like that type also encompasses the non-draft type

packages/noya-state/src/reducers/symbolsReducer.ts
```diff
@@ -167,6 +173,97 @@ export function symbolsReducer(
...
+            elements.shift();
+
+            symbol.overrideValues[index].overrideName =
+              symbolMaster + '/' + elements.join('/');
+            symbol.overrideValues[index].value = value;
+            return;
+          }
+
+          switch (value) {
+            case '': {
```
dabbott, line 248: We should probably pass around `undefined` instead of `''`, even if for the Select element we need `''`

If this only has 2 cases it can be an if/else

packages/app/src/components/inspector/SymbolInstanceOverridesRow.tsx
```diff
@@ -0,0 +1,314 @@
...
+      id="symbol-instance-source"
+      value={symbolId}
+      options={symbolSourceOptions}
+      getTitle={getSymbolMasterTitle}
+      onChange={onChange}
+    />
+  );
+};
+
+function getOverrideElements(
```
dabbott, line 121: I wonder if there's some aspect of this we can share with the SymbolMasterInspector 🤔

dabbott (commented):
This looks great so far. I left a couple misc comments. I know you're still working on this so I didn't look too thoroughly yet

packages/app/src/components/inspector/SymbolInstanceOverridesRow.tsx
milg15: I'm checking this 🤔

packages/app/src/components/inspector/SymbolInstanceOverridesRow.tsx
```diff
@@ -0,0 +1,336 @@
...
```
dabbott, line 130: Since we're already passing in the entire `state` object, I don't think we need to pass `selectors`. Instead, we can  call these first thing within `getOverrideElements`:

```ts
  const themeStyles = getSharedStyles(state);
  const textStyles = getSharedTextStyles(state);
  const symbols = getSymbols(state);
```

packages/app/src/containers/SymbolInstanceInspector.tsx
```diff
@@ -0,0 +1,73 @@
...
+        [dispatch, selectedSymbolIntance.do_objectID],
+      )}
+      onEditSource={useCallback(
+        () => dispatch('goToSymbolSource', selectedSymbolIntance.symbolID),
+        [dispatch, selectedSymbolIntance.symbolID],
+      )}
+    />,
+    symbolMaster && (
+      <SymbolInstanceOverridesRow
+        canOverride={symbolMaster.allowsOverrides}
```
dabbott, line 63: We probably don't need `canOverride` as a prop, since we're already passing the whole `symbolMaster`

dabbott (changes requested):
Overall this looks great and works really well 💯. I found a bunch of small things I think we can improve. I also committed some improvements to the `overrides.ts` utils directly to this branch, so that we can avoid some `as` assertions

packages/noya-state/src/reducers/symbolsReducer.ts
```diff
@@ -167,6 +174,83 @@ export function symbolsReducer(
...
     }
+    case 'setInstanceSymbolSource': {
+      const [, symbolId] = action;
+
+      const symbolMaster = getSymbols(state).find(
+        (symbol) => symbol.symbolID === symbolId,
+      );
+
+      return produce(state, (draft) => {
+        const symbols = getSelectedLayers(draft) as Sketch.SymbolInstance[];
```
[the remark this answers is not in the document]
milg15: Oh this is a good solution! I'll check if I see any other cases

packages/app/src/components/inspector/InspectorPrimitives.tsx
```diff
@@ -19,12 +19,15 @@ export const Title = styled.div(({ theme }) => ({
   userSelect: 'none',
 }));
 
-export const Row = styled.div(({ theme }) => ({
-  flex: '1',
-  display: 'flex',
-  flexDirection: 'row',
-  alignItems: 'center',
-}));
+export const Row = styled.div<{ space?: boolean }>(
```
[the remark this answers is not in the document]
milg15: ohhh! Okey! I didn't know about that one

packages/app/src/components/inspector/SymbolInstanceOverridesRow.tsx
milg15: Isn't it better to only call it once? Instead of calling it every time the getOverrideElements happens 🤔 Oh it doesn't matter?

dabbott: It is better in terms of perf, but for most function calls (especially the first 2 here) it's going to be extremely minor. I'd go with the simpler code and wait till we actually see it causing issues before optimizing here

packages/app/src/components/inspector/SymbolInstanceOverridesRow.tsx
```diff
@@ -0,0 +1,330 @@
...
+};
+
+const SymbolMasterSelector = ({
+  symbols,
+  symbolId,
+  onChange,
+}: {
+  symbols: Sketch.SymbolMaster[];
+  symbolId: string;
+  onChange: (value: string) => void;
```
dabbott, line 99: Should we pass `undefined` in the `"none"` case? Are we assigning `"none"` as the override value currently? (that would work, but probably better to use `undefined`)

packages/noya-state/src/selectors/pageSelectors.ts
```diff
@@ -12,7 +13,9 @@ export type PageMetadata = {
   scrollOrigin: Point;
 };
 
-export const getCurrentPageIndex = (state: ApplicationState) => {
+export const getCurrentPageIndex = (
+  state: ApplicationState | Draft<ApplicationState>,
```
dabbott, line 17: `getCurrentPageIndex` and `getCurrentPage` can probably just use `Draft<ApplicationState>`

dabbott (approved):
Awesome! 🔥 I found 2 tiny things but I'll approve it now

packages/app/src/components/inspector/SymbolInstanceOverridesRow.tsx
milg15: i believe this has to be a string 🤔

milg15: I have this as the call for this function. 

```ts
dispatch(
  'setOverrideTextValue',
  overrideName,
  value === '' || value === 'none' ? undefined : value,
);
```

milg15:
Ohh!! I won't merge it yet I just changed the name of all the over ride reducer from 
```ts
case 'setOverrideValue':
case 'setOverrideTextStyle':
case 'setOverrideSymbolId':
case 'setOverrideThemeStyle':
```

to only one `case 'setOverrideValue'`. I had one for each because I thought they would behave differently but they all do the same and there isn't any difference so I think it's better to keep only one but I'll wait for confirmation on this one.

dabbott:
Good call! That's much better 😄

PR #93, opened 2021-05-29
Symbols tint

milg15:
Basically reuses the fillInspector to work with tints and I also added some correction where selectin 'No Symbol' in the Override wasn't working

image
image

### DIFF
--- a/packages/app/src/containers/FillInspector.tsx
+++ b/packages/app/src/containers/FillInspector.tsx
@@ -9,7 +9,13 @@ import {
 } from '../contexts/ApplicationStateContext';
 import useShallowArray from '../hooks/useShallowArray';
 
-export default memo(function FillInspector() {
+export default memo(function FillInspector({
+  title = 'Fill',
+  allowMoreThanOne = true,
+}: {
+  title?: string;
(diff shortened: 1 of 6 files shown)


packages/app/src/containers/FillInspector.tsx
```diff
@@ -9,7 +9,13 @@ import {
 } from '../contexts/ApplicationStateContext';
 import useShallowArray from '../hooks/useShallowArray';
 
-export default memo(function FillInspector() {
+export default memo(function FillInspector({
+  title = 'Fill',
```
dabbott, line 13: Let's remove the default values and pass them into both `<FillInspector />`s. I think it'll be clearer to be explicit here

dabbott (approved):
👍 Looks great. So the tint is stored as a fill in the sketch file data 🤔. Interesting!

packages/app/src/containers/FillInspector.tsx
milg15: okey!

milg15:
yes I thought it would be somewhere else but It was just there 😄

PR #96, opened 2021-05-30
Export UI part in the inspector

milg15:


packages/app/src/components/inspector/ExportFormatsRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  const isPrefix =
+    exportFormat.namingScheme ===
+    Sketch.ExportFormatNamingScheme.Prefix.valueOf();
+
+  const elements = [
+    <InputField.Root id={scaleInputId} size={55}>
+      <InputField.Input
+        disabled={!canChangeScale}
+        placeholder={canChangeScale ? '' : scale}
```
dabbott, line 83: Is this placeholder so the input text is greyed out when the field is disabled? Typically we'd do that using the disabled psuedoclass, e.g. within our `InputElement`:

```ts
'&:disabled': {
    color: theme.colors.textDisabled,
},
```

packages/noya-designsystem/src/components/InputField.tsx
```diff
@@ -1,59 +1,108 @@
...
+/* ----------------------------------------------------------------------------
+ * Dropdown
+ * ------------------------------------------------------------------------- */
+
+const DropdownContainer = styled.span(({ theme }) => ({
+  position: 'absolute',
+  right: 0,
+}));
+
+interface InputFieldDropdownProps {
```
dabbott, line 83: Here we should try to reuse the generic `items` and `onSelect` props we use in `DropdownMenu` and `ContextMenu`, rather than creating a new pattern

https://github.com/noya-app/noya/blob/5380076b51cc8e225b854cc9fecbd7f2ae751450/packages/noya-designsystem/src/components/DropdownMenu.tsx#L38-L39

packages/noya-designsystem/src/components/InputField.tsx
```diff
@@ -1,59 +1,108 @@
...
-    <LabelContainer labelPosition={labelPosition}>{children}</LabelContainer>
+    <DropdownContainer>
+      <NoyaDropdownMenu.Root<string>
+        items={list.map((e) => ({
+          value: e.value.toString(),
+          title: e.title,
+        }))}
+        onSelect={onSumbitList}
+      >
+        <Button id={'a'} variant="thin">
```
dabbott, line 101: `id` will need to be a prop, since this element is in a list, and every HTML id must be globally unique

packages/noya-designsystem/src/components/InputField.tsx
```diff
@@ -200,10 +260,14 @@ function InputFieldRoot({
   labelPosition = 'end',
   labelSize = 6,
 }: InputFieldRootProps) {
-  const contextValue = useMemo(() => ({ labelPosition, labelSize }), [
-    labelPosition,
-    labelSize,
-  ]);
+  const hasDropdown = Children.toArray(children).some(
```
dabbott, line 263: I added this. We can detect `hasDropdown` automatically by checking the types of the `children` elements, which is a little more convenient

packages/noya-state/src/reducers/applicationReducer.ts
```diff
@@ -224,6 +226,13 @@ export function applicationReducer(
     case 'goToSymbolSource':
     case 'setOverrideValue':
       return symbolsReducer(state, action);
+    case 'setScale':
```
dabbott, line 229: We might want to call these

`setExportScale`
`setExportName`
`setExportFileFormat`
`setExportNamingScheme`

Since we have a lot of other things that `setName` and even the others could refer to

packages/noya-state/src/reducers/exportReducer.ts
```diff
@@ -0,0 +1,113 @@
...
+      });
+    }
+    case 'setName': {
+      const [, index, value] = action;
+
+      return produce(state, (draft) => {
+        const layers = getSelectedLayers(draft);
+
+        layers.forEach(
+          (layer) => (layer.exportOptions.exportFormats[index].name = value),
```
dabbott, line 82: I should add an eslint rule for this, but avoid "assignment expressions", where we assign a value as part of an expression. I.e. add curly braces here

packages/noya-state/src/reducers/exportReducer.ts
```diff
@@ -0,0 +1,113 @@
...
+  | [
+      type: 'setNamingScheme',
+      index: number,
+      value: Sketch.ExportFormatNamingScheme,
+    ]
+  | [type: 'addExportFormat']
+  | [type: 'deleteExportFormat', index: number];
+
+export function exportReducer(
+  state: ApplicationState,
```
dabbott, line 19: Can we pass just the `ExportOptions` type here? That'll make everything a little simpler I think. We can call the reducer from within `applicationReducer` for each layer.

Similar to https://github.com/noya-app/noya/blob/5380076b51cc8e225b854cc9fecbd7f2ae751450/packages/noya-state/src/reducers/applicationReducer.ts#L147-L158

packages/app/src/containers/ExportInspector.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <ArrayController<FileFormat.ExportFormat>
+      title={title}
+      id={title}
+      key={title}
+      value={exportFormats}
+      onClickPlus={useCallback(() => {
+        dispatch('addExportFormat');
+      }, [dispatch])}
+    >
+      {({ item, index }: { item: FileFormat.ExportFormat; index: number }) => (
```
dabbott, line 41: We'll need a `useCallback` around this function to avoid re-renders

packages/app/src/containers/ExportInspector.tsx
```diff
@@ -0,0 +1,68 @@
...
+import type FileFormat from '@sketch-hq/sketch-file-format-ts';
+import ArrayController from '../components/inspector/ExportArrayController';
+import ExportFormatsRow from '../components/inspector/ExportFormatsRow';
+import ExportPreviewRow from '../components/inspector/ExportPreviewRow';
+
+export default memo(function ExportInspector() {
+  const title = 'Exports';
+  const dispatch = useDispatch();
+
+  const selectedLayer = useShallowArray(
```
dabbott, line 17: We don't need `useShallowArray` here, since we're not passing this to the dependencies array of any hooks

packages/app/src/containers/ExportInspector.tsx
```diff
@@ -0,0 +1,68 @@
...
+    useSelector(Selectors.getSelectedLayers),
+  )[0] as Sketch.SymbolInstance;
+
+  const symbolMaster = useShallowArray(useSelector(Selectors.getSymbols)).find(
```
dabbott, line 21: Same here, no need for `useShallowArray`

packages/app/src/components/inspector/ExportFormatsRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+    [scaleInputId, namingSchemeInputId, fileFormatInputId],
+  );
+
+  const canChangeScale =
+    exportFormat.fileFormat === Sketch.ExportFileFormat.JPG ||
+    exportFormat.fileFormat === Sketch.ExportFileFormat.PNG;
+
+  const isPrefix =
+    exportFormat.namingScheme ===
+    Sketch.ExportFormatNamingScheme.Prefix.valueOf();
```
dabbott, line 77: What does `valueOf` do here?

packages/app/src/components/inspector/ExportFormatsRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+        case fileFormatInputId:
+          return <Label.Label>Format</Label.Label>;
+        default:
+          return null;
+      }
+    },
+    [scaleInputId, namingSchemeInputId, fileFormatInputId],
+  );
+
+  const canChangeScale =
```
dabbott, line 71: Sketch allows scaling on every format except `SVG`, so we should do the same

packages/noya-state/src/reducers/exportReducer.ts
```diff
@@ -0,0 +1,113 @@
...
+        const number = isNaN(parseFloat(value))
+          ? parseFloat(value.slice(0, -1))
+          : parseFloat(value);
+
+        const visibleScale =
+          value.slice(-1) === 'w'
+            ? Sketch.VisibleScaleType.Width
+            : value.slice(-1) === 'h'
+            ? Sketch.VisibleScaleType.Height
+            : Sketch.VisibleScaleType.Scale;
```
dabbott, line 38: We should only be parsing/formatting this once. We currently have similar code both here and in the UI component. I think we could make a function that parses the input and returns an object, which we can then pass as the action `value`

```ts
type ExportScale = { scale: number, visibleScaleType: Sketch.VisibleScaleType }

function parseScale(scaleText: string): ExportScale | undefined {
  // ...
}
```

If the user types something invalid, then `parseScale` should return `undefined`, and we call `reset` on the input. It should never dispatch an action.

dabbott (changes requested):
This works well! 

I think we may need to use a different value for the Size field in the case of `w` or `h` units? Maybe `absoluteSize`, but I'm not sure 

| Noya | Sketch |
| --- | --- |
| Screen Shot 2021-05-31 at 12 48 43 PM | Screen Shot 2021-05-31 at 12 47 46 PM |

packages/app/src/components/inspector/ExportFormatsRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+        )}
+      />
+      <InputField.Label>...</InputField.Label>
+    </InputField.Root>,
+    <InputField.Root size={65} id={fileFormatInputId}>
+      <Select
+        id={'fileFormat-select'}
+        value={exportFormat.fileFormat}
+        options={fileFormatOptions}
+        getTitle={(id) => id.toUpperCase()}
```
dabbott, line 128: `useCallback`

packages/app/src/components/inspector/ExportFormatsRow.tsx
dabbott: Actually, even better:

```ts
color: disabled ? theme.colors.textDisabled : theme.colors.text,
```

packages/noya-state/src/reducers/exportReducer.ts
milg15: Do add this in a selector ? 🤔

dabbott: I'd probably make a new file `noya-state/src/exportOptions.ts` and put it in there. There's a lot of stuff in selectors that we can probably move elsewhere at some point

packages/noya-designsystem/src/components/InputField.tsx
milg15: I was using numbers before so it didn't work with it but I changed it a bit so I send strings instead 👍

packages/noya-designsystem/src/components/InputField.tsx
```diff
@@ -64,35 +112,62 @@ function InputFieldLabel({ children = false }: InputFieldLabelProps) {
...
+    border: '0',
+    outline: 'none',
+    minWidth: '0',
+    textAlign: textAlign ?? 'left',
+    alignSelf: 'stretch',
+    borderRadius: '4px',
+    paddingTop: '4px',
+    paddingBottom: '4px',
+    paddingLeft:
+      labelPosition === 'start'
```
dabbott, line 143: If we do `hasLabel ?` first, we can probably simplify these a little

```ts
paddingLeft: hasLabel && labelPosition === 'start' ? `${6 + labelSize + 6}px` : '6px'
```

Maybe paddingRight can be simpler too

packages/noya-designsystem/src/components/InputField.tsx
```diff
@@ -200,10 +276,17 @@ function InputFieldRoot({
   labelPosition = 'end',
   labelSize = 6,
 }: InputFieldRootProps) {
-  const contextValue = useMemo(() => ({ labelPosition, labelSize }), [
-    labelPosition,
-    labelSize,
-  ]);
+  const hasDropdown = Children.toArray(children).some(
```
dabbott, line 279: Very minor, but both of these can share the same `toArray` call. E.g. 

```ts
const childrenArray = Children.toArray(children)

const hasDropdown = childrenArray.some ...
const hasLabel = ...
```

packages/noya-state/src/exportOptions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+type ExportScale = {
+  scale: number;
+  absoluteSize: number;
+  visibleScaleType: Sketch.VisibleScaleType;
+};
+
+export function parseScale(
+  scaleText: string,
+  rect: Sketch.Rect,
```
dabbott, line 11: I didn't really realize before, but if we include the rect here, then we're going to have to pass it all the way in to our input validation, and it really isn't necessary there. I think we're better off parsing & validating just the user input `[size, visibleScaleType]` here, and handling the scale/rect calculation in the reducer

packages/app/src/components/inspector/ExportFormatsRow.tsx
```diff
@@ -0,0 +1,157 @@
...
+      labelPosition={isPrefix ? 'end' : 'start'}
+    >
+      <InputField.Input
+        value={exportFormat.name}
+        placeholder={isPrefix ? 'Prefix' : 'Suffix'}
+        onSubmit={onChangeName}
+        textAlign={isPrefix ? 'end' : 'start'}
+      />
+      <InputField.DropdownMenu
+        list={[
```
dabbott, line 116: `useMemo`

dabbott (changes requested):
These components that render lists of input fields can be pretty slow, so I left some more feedback about how we can improve the performance here

packages/app/src/containers/ExportInspector.tsx
```diff
@@ -0,0 +1,85 @@
...
+      value={exportFormats}
+      onClickPlus={useCallback(() => {
+        dispatch('addExportFormat');
+      }, [dispatch])}
+    >
+      {({ item, index }: { item: FileFormat.ExportFormat; index: number }) => (
+        <ExportFormatsRow
+          id={`exportFormat-${index}}`}
+          last={index === exportFormats.length - 1}
+          frame={selectedLayer.frame}
```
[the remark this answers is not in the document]
milg15: 🤔 Yes but I'll have to do it in the application reducer since I am only sending the exportOptions to the reducer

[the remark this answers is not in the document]
milg15: I added it to the application reducer 😆 Can you validate if this is a better approach

packages/noya-designsystem/src/components/InputField.tsx
```diff
@@ -1,59 +1,107 @@
...
+ * Dropdown
+ * ------------------------------------------------------------------------- */
+
+const DropdownContainer = styled.span(({ theme }) => ({
+  position: 'absolute',
+  right: 0,
+}));
+
+interface InputFieldDropdownProps<T extends string> {
+  list: MenuItem<T>[];
```
dabbott, line 87: Can we call this `items` to be consistent with the Dropdown/ContextMenu components?

packages/noya-designsystem/src/components/InputField.tsx
```diff
@@ -64,35 +112,60 @@ function InputFieldLabel({ children = false }: InputFieldLabelProps) {
...
+    minWidth: '0',
+    textAlign: textAlign ?? 'left',
+    alignSelf: 'stretch',
+    borderRadius: '4px',
+    paddingTop: '4px',
+    paddingBottom: '4px',
+    paddingLeft:
+      hasLabel && labelPosition === 'start' ? `${6 + labelSize + 6}px` : '6px',
+    paddingRight:
+      labelPosition === 'start'
```
dabbott, line 145: We probably want to check `hasLabel` here first to get the correct behavior in the case of a dropdown with a `labelPosition` of `start`. Currently this would short-circuit and do `6px` even if there were a dropdown

packages/noya-state/src/exportOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const absoluteSize =
+    visibleScaleType === Sketch.VisibleScaleType.Scale ? 0 : scaleValue;
+
+  const scale =
+    visibleScaleType === Sketch.VisibleScaleType.Scale
+      ? scaleValue
+      : absoluteSize;
+
+  return { scale, absoluteSize, visibleScaleType };
```
dabbott, line 31: Since `scale` may not be accurate here, it might be more accurate to return `{ size, visibleScaleType }`, and figure out absoluteSize vs. scale later

packages/noya-state/src/exportOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export function getScaleUnits(visibleScaleType: Sketch.VisibleScaleType) {
+  return visibleScaleType !== Sketch.VisibleScaleType.Scale
```
dabbott, line 35: Can this be a `switch`?

packages/noya-state/src/reducers/applicationReducer.ts
```diff
@@ -224,6 +226,32 @@ export function applicationReducer(
...
+    case 'setExportFileFormat':
+    case 'setExportNamingScheme':
+    case 'addExportFormat':
+    case 'deleteExportFormat':
+      const pageIndex = getCurrentPageIndex(state);
+      const layerIndexPaths = getSelectedLayerIndexPaths(state);
+
+      return produce(state, (draft) => {
+        accessPageLayers(draft, pageIndex, layerIndexPaths).forEach((layer) => {
+          layer.exportOptions = exportReducer(layer.exportOptions, action);
```
dabbott, line 240: Maybe we could pass the `layer.frame` into `exportReducer` as a 3rd parameter? We don't use that pattern elsewhere yet, but we'll probably have other cases like this where we need to send more data into a reducer. Then we can handle `setExportScale` entirely in the reducer

dabbott (approved):

packages/noya-designsystem/src/components/InputField.tsx
milg15:
```css
paddingRight:
      (!hasLabel || !hasDropdown) && labelPosition === 'end'
        ? '6px'
        : `${6 + (hasLabel ? labelSize + 6 : 0) + (hasDropdown ? 11 : 0)}px`,
``` 

Like this ? 🤔

packages/noya-state/src/exportOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+  size: number;
+  visibleScaleType: Sketch.VisibleScaleType;
+};
+
+export function parseScale(scaleText: string): ExportSize | undefined {
+  const size = isNaN(parseFloat(scaleText))
+    ? parseFloat(scaleText.slice(0, -1))
+    : parseFloat(scaleText);
+
+  if (isNaN(size) && size > 0) return undefined;
```
dabbott, line 19: Just noticed this needs to be `||` and slightly different logic, cause a number can't be both NaN and greater than 0. I'll update it after I merge

packages/noya-designsystem/src/components/InputField.tsx
dabbott: I'm not sure 😂. I might try to simplify it after merging

dabbott:
LGTM 🚀

packages/noya-designsystem/src/components/InputField.tsx
dabbott: I'm going to update it to this

Screen Shot 2021-06-02 at 1 24 40 PM

packages/noya-state/src/exportOptions.ts
milg15: 👁️ I didn't notice tht one !

PR #100, opened 2021-06-03
Swatches from Color Inspector

milg15:
I am still figuring out the best way to change the SwatchId of the fill.

### DIFF
--- a/packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
+++ b/packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
@@ -17,6 +17,7 @@ import { Selectors } from 'noya-state';
 import { memo, useCallback, useMemo, useState } from 'react';
 import styled from 'styled-components';
 import ColorInspector from './ColorInspector';
+import { uuid } from 'noya-renderer';
 
 const Content = styled(Popover.Content)(({ theme }) => ({
   width: '240px',
@@ -152,7 +153,6 @@ export default memo(function ColorInputFieldWithPicker({
   const values = useMemo(() => [value], [value]);
 
   const [state, dispatch] = useApplicationState();
-
   const [swatchLayout, setSwatchLayout] = useState<SwatchLayout>('grid');
 
   const sharedSwatches = Selectors.getSharedSwatches(state);
@@ -178,18 +178,17 @@ export default memo(function ColorInputFieldWithPicker({
             id={'handle-theme-color'}
             full={true}
             onClick={useCallback(() => {
-              if (isSwatch) {
-                onChange({
-                  ...value,
-                  swatchID: undefined,
-                });
-                return;
-              }
+              const id = uuid();
+              onChange({
+                ...value,
+                swatchID: isSwatch ? undefined : id,
+              });
+              if (isSwatch) return;
 
-              const swatchName = prompt('Swatch Name');
+              const swatchName = prompt('New Swatch Name');
               if (!swatchName) return;
 
-              dispatch('addSwatch', swatchName, value);
+              dispatch('addSwatch', swatchName, value, id);
             }, [onChange, isSwatch, dispatch, value])}
           >
             {isSwatch ? 'Detach Theme Color' : 'Create Theme Color'}
--- a/packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
+++ b/packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
@@ -35,6 +35,8 @@ const StyledArrow = styled(Popover.Arrow)(({ theme }) => ({
 
 const PaddedSection = styled.section({
   padding: '10px',
+  display: 'flex',
+  flexDirection: 'column',
 });
 
 const Row = styled.div(({ theme }) => ({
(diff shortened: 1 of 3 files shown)


packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
```diff
@@ -150,12 +152,15 @@ export default memo(function ColorInputFieldWithPicker({
...
-  const [state] = useApplicationState();
-
+  const [state, dispatch] = useApplicationState();
   const [swatchLayout, setSwatchLayout] = useState<SwatchLayout>('grid');
 
   const sharedSwatches = Selectors.getSharedSwatches(state);
 
+  const isSwatch =
+    value.swatchID &&
+    sharedSwatches.some((e) => e.do_objectID === value.swatchID);
```
dabbott, line 162: I think theoretically we shouldn't need to check `sharedSwatches` at all, cause all `swatchID` should be valid if they exist? Did you find this was needed?

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
```diff
@@ -168,6 +173,26 @@ export default memo(function ColorInputFieldWithPicker({
...
+            onClick={useCallback(() => {
+              const id = uuid();
+              onChange({
+                ...value,
+                swatchID: isSwatch ? undefined : id,
+              });
+              if (isSwatch) return;
+
+              const swatchName = prompt('New Swatch Name');
+              if (!swatchName) return;
```
dabbott, line 189: We should call `onChange` after we know we have a swatchName, so we don't end up in an inconsistent state

packages/noya-designsystem/src/components/Button.tsx
```diff
@@ -11,7 +11,8 @@ type ButtonVariant = 'normal' | 'thin';
 const ButtonElement = styled.button<{
   active: boolean;
   variant: ButtonVariant;
-}>(({ theme, active, disabled, variant }) => ({
+  full: boolean;
```
dabbott, line 14: We might need to add this, but if we do, I think we should try to use flexbox.

In this case though, I think we can get the right layout by making `PaddedSection` a flex container:

Screen Shot 2021-06-03 at 11 08 23 AM

dabbott (changes requested):
I tried it and it looks great 💯

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
milg15: Maybe if we delete a swatch but they color still has the swatchId? 🤔

dabbott: In that case, we should probably update our swatch deletion code to traverse layers/styles and delete the swatchIDs. Similar to how we do `visitLayerColors` and `visitStyleColors` with "setSwatchColor"

packages/noya-designsystem/src/components/Button.tsx
```diff
@@ -62,6 +62,7 @@ interface ButtonRootProps {
   disabled?: boolean;
   variant?: ButtonVariant;
   tooltip?: ReactNode;
+  full?: boolean;
```
dabbott, line 65: Remove

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
dabbott: Do you want to update the deletion code in this PR? Or should we do it another time?

dabbott (approved):

milg15:
How update the deletion ?

dabbott:
Currently when we `setSwatchColor`, we update every instance of that color with the same swatchID.

Screen Shot 2021-06-03 at 12 25 34 PM

We need to do the same when removing a swatch. For any color with the same swatchID, we need to set the swatchID of that color to undefined.

Screen Shot 2021-06-03 at 12 26 24 PM

milg15:
Oh! I can add it to this PR actually

packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -505,6 +529,19 @@ export function themeReducer(
...
         draft.sketch.document.sharedSwatches = sharedSwatches;
+
+        layerIndexPathsWithSwatchId.forEach((layerPath) =>
+          accessPageLayers(
+            draft,
+            layerPath.pageIndex,
+            layerPath.indexPaths,
+          ).forEach((layer) => {
+            layer.style?.fills?.forEach((f) => {
+              if (f.color.swatchID && ids.includes(f.color.swatchID))
```
dabbott, line 540: Very minor but we already checked this condition above, so I think it should be safe to just `delete` the swatch without the condition. Anything in `layerIndexPathsWithSwatchId` should have the correct swatchID

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
dabbott: Now that we've updated the swatch deletion, can we delete this `.some` check?

packages/noya-state/src/reducers/themeReducer.ts
milg15: The layer can have multiple fills

dabbott: Oh yeah, good point. Ok, keep it the way it is

PR #102, opened 2021-06-04
Smart snap

### LINKED ISSUE #69: Smart snap
## Feature

Smart snap improves the layer dragging UX by automatically aligning the dragged layer with nearby layers.

### What should snap?

Snapping always occurs between the bounding rect for the selected layers + nearby layer bounding rect. Multiple layers can be selected, but snapping will not occur for individual layers within the selection, only the combined bounding rect of all selected layers (the same behavior as quick measure).

A layer in an artboard should only snap to other layers within that artboard and the artboard itself. It should never snap to a layer outside that artboard. A layer outside all artboards should also not snap to a layer in an artboard. If a selection contains layers with different parents (artboard or top-level), the selection should not snap to layers in any artboard.

Only layers currently visible on the canvas should participate in snapping. I.e. layers that intersect or are contained within the rect of the currently visible portion of the canvas.

### Algorithm

We should think of the x and y axes as **completely independent**. Snapping along one axis will never influence snapping along the other. For that reason, let's just consider the y axis for now, bearing in mind that it will be the same for x also.

Each rect (whether for the selection, or the layer's frame) has 3 points of interest: minY, midY, and maxY. Each of these can snap to any of these for another layer (I'm using the term "point of interest", but maybe "value of interest" would be more accurate, since it's just on a single axis). Here are the possible combinations for 2 non-intersecting layers:

Normal Spec

Snapping should also occur when layers intersect, including when one layer completely contains another. Here are a few non-exhaustive examples of this:

Intersecting Spec

### Snap distance

As the user drags a layer, we should snap that layer's point of interest (_value of interest_) to another point of interest when they're within some `n` distance (distance measured along 1 axis, not distance between 2 points). `n` is a function of the current zoom. In Sketch, `n` seems to be about 6 at 100% zoom and 2 at 200% zoom.

We should always consider all possible snaps and choose the snap with the smallest distance. E.g. suppose we have 2 vertical center points we could snap to:

Multiple snaps spec

In the above case, we'll snap to the teal layer's center because it's closer to the vertical center of the blue circle.

### Drawing guides

We should draw 3 kinds of guides:

- **Alignment guides** - these lines are drawn between the snapping points of interest. If we're snapping on the `y` axis, then we would draw an alignment guide along the `x` axis. The line would go from the minimum x value to the maximum x value along the x axis, when combining the minX and maxX of both rects. I.e. I think it's from `Math.min(rect1.minX, rect2.minX)` to `Math.max(rect1.maxX, rect2.maxX)`
- **Measurement guides** - these lines indicate distances. We should only draw them when the layers are non-intersecting (`Primitives.rectsIntersect(...)`). Measurement guides should always be drawn from the midpoint of the selected rect to the nearest edge of the snapping rect.
- **Extension guides** - these lines extend the edge of a rect so that we can draw a measurement guide to it.

If multiple points of interest happen to align, we should draw alignment guides along all of them. For example, when snapping the vertical centers of 2 layers of the same height, we should see the minY, midY, and maxY alignment guides:

Multiple alignment guides spec

If multiple points of interest _between multiple layers_ happen to align, we should draw alignment guides with all of them. However, we should still only show one measurement guide, for the nearest of these layers. E.g.:

Multiple alignment guides spec

Ideally we can share measurement guide and extension guide logic with the "quick measure" code. I think the main difference is that with smart snap, we only show a measurement guide along an axis if we have a snap along that axis - so we may need to update/refactor `getGuides` for more flexibility.

## Simplifications

We don't need to implement these:

- Sketch has special behavior for equidistant layers, where it shows distances with a red transparent fill
Untitled

leannepepper:
Adding Smart Snap Feature

closes #69


packages/app/src/containers/Canvas.tsx
```diff
@@ -363,13 +379,14 @@ export default memo(function Canvas() {
           const { origin, current } = state.interactionState;
 
           const layers = Selectors.getLayersInRect(
-            CanvasKit,
+            // CanvasKit,
```
leannepepper, line 382: Would you mind if I remove CanvasKit from getLayersInRect?

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -19,8 +29,158 @@ import {
...
+      includeHiddenLayers: false,
+      includeArtboardLayers: true,
+    },
+  );
+  let layerIsInArtboard = false;
+  let artboardLayers: any[] = [];
+
+  allVisibleLayers
+    .filter((layer) => layer._class === 'artboard' || layer._class === 'group')
+    //TODO : why does TS not recognize artboard.layers ?
```
leannepepper, line 75: I couldn't figure out why TS couldn't recognize the type of artboard, any suggestions?

packages/app/src/containers/Canvas.tsx
leannepepper: removing

dabbott (approved):
Perfect code!

PR #105, opened 2021-06-09
gradients-fill-inspector

milg15:
Finally it makes sense! Todo (In this issue or in a new one):
* Sort Gradient Stops by position after changing it. Change the way we find/store the index ?
* Selecting a SwatchColor doesn't work on Gradients 🤔 This might require to move the selection of the index further up into the app components


packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -83,34 +88,42 @@ export function themeReducer(
       });
     }
     case 'addSwatch': {
+      const [, name, color, id] = action;
```
dabbott, line 91: I think we already merged this swatch stuff. Maybe we need to rebase?

packages/noya-state/src/reducers/styleReducer.ts
```diff
@@ -280,6 +289,49 @@ export function styleReducer(
...
+      return produce(state, (draft) => {
+        if (!draft.fills || !draft.fills[index]) return;
+        draft.fills[index].fillType = type;
+
+        if (!draft.fills[index].gradient) {
+          draft.fills[index].gradient = Models.fill.gradient;
+        }
+      });
+    }
+    case 'setGradientColor': {
```
dabbott, line 304: We'll probably want a separate `setGradientColor` and `setGradientPosition`, for more granular control. This should help with the jankiness when dragging

packages/noya-state/src/reducers/styleReducer.ts
```diff
@@ -280,6 +289,49 @@ export function styleReducer(
...
+        if (!draft.fills || !draft.fills[fillIndex]) return;
+        if (!draft.fills[fillIndex].gradient) return;
+
+        if (gradientIndex !== -1) {
+          draft.fills[fillIndex].gradient.stops[
+            gradientIndex
+          ].position = position;
+          draft.fills[fillIndex].gradient.stops[gradientIndex].color = color;
+        } else {
+          draft.fills[fillIndex].gradient.stops.push({
```
dabbott, line 317: Maybe should sort the gradient stops whenever we push or modify their position?

packages/noya-state/src/reducers/styleReducer.ts
```diff
@@ -280,6 +289,49 @@ export function styleReducer(
...
+        if (!draft.fills || !draft.fills[index]) return;
+        draft.fills[index].fillType = type;
+
+        if (!draft.fills[index].gradient) {
+          draft.fills[index].gradient = Models.fill.gradient;
+        }
+      });
+    }
+    case 'setGradientColor': {
+      const [, fillIndex, gradientIndex, position, color] = action;
```
dabbott, line 305: Maybe `gradientIndex` => `stopIndex`?

packages/noya-designsystem/src/components/ColorPicker.tsx
```diff
@@ -23,14 +24,23 @@ const colorModel: ColorModel<RgbaColor> = {
 
 interface Props {
   value: Sketch.Color;
-  onChange: (color: Sketch.Color) => void;
+  gradients?: Sketch.GradientStop[];
```
dabbott, line 27: I think this component would be simpler if we had a separate version specifically for gradients. E.g. `ColorPicker.tsx`, `GradientPicker.tsx`. I think the only component where we should combine everything into one is `ColorInputFieldWithPicker`, for the popup to work correctly

packages/noya-designsystem/src/components/ColorInputField.tsx
```diff
@@ -1,30 +1,39 @@
...
+    '&:focus': {
+      boxShadow: `0 0 0 1px ${theme.colors.sidebar.background}, 0 0 0 3px ${theme.colors.primary}`,
+    },
+  }),
+);
 
 interface Props {
   id?: string;
-  value: Sketch.Color;
+  value: Sketch.Color | Sketch.Gradient;
```
dabbott, line 23: Can we do separate components here? `ColorInputField`, `GradientInputField`

packages/noya-colorpicker/src/types.ts
```diff
@@ -42,5 +48,9 @@ export interface ColorModel<T extends AnyColor> {
 export interface ColorPickerBaseProps<T extends AnyColor> {
   className: string;
   color: T;
-  onChange: (newColor: T) => void;
+  onChange: (newColor: T, index?: number, position?: number) => void;
```
dabbott, line 51: We should have a separate component for color vs. gradient

packages/noya-colorpicker/src/hooks/useEventCallback.ts
```diff
@@ -1,14 +1,20 @@
-import { useRef, useEffect, useCallback } from "react";
+import { useRef, useEffect, useCallback } from 'react';
 
 // Saves incoming handler to the ref in order to avoid "useCallback hell"
-function useEventCallback<T>(handler?: (value: T) => void): (value: T) => void {
+function useEventCallback<T>(
```
dabbott, line 4: I think we can use generics to handle both cases:

```ts
function useEventCallback<T extends any[]>(handler?: (...values: T) => void): (...values: T) => void {
  const callbackRef = useRef(handler);

  useEffect(() => {
    callbackRef.current = handler;
  });

  return useCallback((...values: T) => callbackRef.current && callbackRef.current(...values), []);
}
```

If we remove the type annotations from `useEventCallback<string>(...)` => `useEventCallback(...)` I think they should be inferred correctly. We can do `useEventCallback<[string]>(...)` if we still need type annotations in some cases

dabbott (commented):
I think this is looking pretty good! My main feedback is: 

- I think we should mostly still try to split up color handling and gradient handling into separate components/types, even if we don't split up `ColorInputFieldWithPicker` due to the popup issue
- We'll need some way to focus a specific `Pointer` on click 🤔. Maybe in the case of gradients we'll want separate handling for the pointers and the track behind them

packages/noya-designsystem/src/components/ColorPicker.tsx
milg15: Ohh!! Okey I'll try separating them 👍

packages/noya-designsystem/src/components/ColorInputField.tsx
milg15: yes! They look almost the same 🤔 That's why I combined them but I'll separate them

packages/noya-state/src/reducers/styleReducer.ts
milg15: I have an issue with this 😅 I did used to do this but since I am storing the index of the current stop I am modifying when I resort the array this index changes but I am not 100% sure where in the code I should do this index change 🤔

dabbott: That's a good point, I didn't think of that. Maybe the way you have it now is good then, although I will probably need to add a sort to the rendering code at some point

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
```diff
@@ -145,49 +168,166 @@ export default memo(function ColorInputFieldWithPicker({
...
+        />
       </Popover.Trigger>
       <Content side="bottom" align="center">
         <PaddedSection>
-          <ColorInspector
-            id={`${id}-panel`}
-            colors={values}
-            onChangeColor={onChange}
-          />
+          <Row>
```
dabbott, line 256: I know we need the `Popover.Root` and `Popover.Trigger` to stay mounted the whole time, but I wonder if we could split everything else in this `<Content>` element into 2 components? One for colors and one for gradients? I think it would help keep things organized. We might be able to move some of the logic above, e.g. `const fillOption = ...` into each component. We could leave the new components in this same file.

E.g. `ColorPickerContent`, `GradientPickerContent`

packages/app/src/containers/BorderInspector.tsx
```diff
@@ -57,7 +57,7 @@ export default memo(function BorderInspector() {
         }) => (
           <BorderRow
             id={`border-${index}`}
-            color={item.color}
+            value={item.fillType === 1 ? item.gradient : item.color}
```
dabbott, line 60: Can we use a Sketch constant here instead of 1?

packages/app/src/containers/FillInspector.tsx
```diff
@@ -66,8 +65,34 @@ export default memo(function FillInspector({
         }) => (
           <FillRow
             id={`fill-${index}`}
-            color={item.color}
+            value={
+              item.fillType === 1
```
dabbott, line 69: Can we use Sketch constants here?

packages/noya-colorpicker/src/components/Gradient.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { useColorPicker } from '../contexts/ColorPickerContext';
+import { rgbaToHsva } from '../utils/convert';
+import { colorAt, RGBA_MAX } from '../utils/colorAt';
+import { RgbaColor } from '../types';
+import { Interaction, Interactive } from './GradientSlider';
+import Pointer from './Pointer';
+
+
+const Container = styled.div<{ background: string }>(({ background }) => ({
+  position: 'relative' as any,
```
dabbott, line 15: `as any`? 🤔

packages/noya-colorpicker/src/components/GradientSlider.tsx
```diff
@@ -0,0 +1,204 @@
...
+interface Props {
+  onMove: (interaction: Interaction) => void;
+  onMoveChange: (moving: boolean) => void;
+  onClick: (interaction: Interaction) => void;
+  onKey: (offset: Interaction) => void;
+  isOnPoint: (interaction: Interaction) => boolean;
+  children: React.ReactNode;
+}
+
+const InteractiveBase = ({
```
dabbott, line 54: Since this file is kinda complex and still mostly the same as `Interactive`, it'd be nice if we could share some/all of the logic. What's unique about this version? I'm wondering if we can make it an optional prop?

packages/noya-colorpicker/src/components/GradientSlider.tsx
```diff
@@ -0,0 +1,204 @@
...
+    [onMoveCallback],
+  );
+
+  const handleMoveStart = useCallback(
+    ({ nativeEvent: event }: React.MouseEvent | React.TouchEvent) => {
+      event.preventDefault();
+
+      if (!isValid(event)) return;
+
+      if (!isOnPoint(getRelativePosition(container.current!, event))) {
```
dabbott, line 111: We could probably use `event.target` to determine if we're clicking a pointer. E.g. we could add a `className="pointer"` to the pointer elements, and then:

```ts
if (event.target instanceof HTMLElement && event.target.classList.contains('pointer')) {
  onClickCallback(getRelativePosition(container.current!, event));
}
```

packages/noya-colorpicker/src/types.ts
dabbott: Are these still used? Can we revert to what they were before?

packages/noya-designsystem/src/components/ColorInputField.tsx
```diff
@@ -1,30 +1,48 @@
...
 }
 
 export default forwardRef(function ColorInputField(
   { id, value, ...rest }: Props,
   ref: ForwardedRef<HTMLButtonElement>,
 ) {
-  const colorString = useMemo(() => sketchColorToRgbaString(value), [value]);
+  const background = useMemo(
+    () =>
+      value._class === 'color'
```
dabbott, line 39: Maybe a `switch`?

packages/noya-designsystem/src/utils/getGradientBackground.tsx
```diff
@@ -0,0 +1,19 @@
...
+) {
+  const color = [...value]
+    .sort((a, b) => a.position - b.position)
+    .map((g) => `${sketchColorToRgbaString(g.color)} ${g.position * 100}%`)
+    .join(', \n');
+
+  const position =
+    type === Sketch.GradientType.Radial
+      ? 'radial-gradient(circle'
+      : `linear-gradient(${type !== undefined ? 0 : 90}deg`;
```
dabbott, line 16: Why does an `undefined` type change the direction?

packages/noya-designsystem/src/utils/getGradientBackground.tsx
```diff
@@ -0,0 +1,19 @@
...
+export function getGradientBackground(
+  value: Sketch.GradientStop[],
+  type?: Sketch.GradientType,
+) {
+  const color = [...value]
+    .sort((a, b) => a.position - b.position)
+    .map((g) => `${sketchColorToRgbaString(g.color)} ${g.position * 100}%`)
+    .join(', \n');
+
+  const position =
```
dabbott, line 13: Can we use a `switch` here?

```ts
switch (type) {
  case Sketch.GradientType.Radial: 
    return `radial-gradient(circle, ${color})`
}
```

dabbott: Can we add angular/conic gradient as an option? https://developer.mozilla.org/en-US/docs/Web/CSS/conic-gradient()

If it's hard, we can come back to it

packages/noya-state/src/reducers/styleReducer.ts
```diff
@@ -280,6 +308,97 @@ export function styleReducer(
...
+    case 'setFillGradientPosition':
+    case 'setBorderGradientPosition': {
+      const [, index, stopIndex, position] = action;
+      const category =
+        action[0] === 'setFillGradientPosition' ? 'fills' : 'borders';
+
+      return produce(state, (draft) => {
+        if (
+          !draft[category] ||
+          !draft[category]![index] ||
```
dabbott, line 353: We typically want to avoid using `!` to ignore null/undefined values, since it's much easier to cause a crash. If we can't get the types to work out, we should make 2 separate `case`s.

Separately, maybe we could make a `gradientReducer`? And then pass either fill or border gradient into it? I think that'll result in shorter code still.

packages/noya-colorpicker/src/utils/colorAt.ts
```diff
@@ -0,0 +1,61 @@
...
+    let k: keyof RgbaColor;
+    for (k in start) {
+        color[k] = step[k] * i + start[k];
+        color[k] = color[k] < 0 ? color[k] + max[k] : color[k];
+    }
+
+    return color;
+  }
+
+  export function colorAt(stops: PositionHsva[], pos: number, max: RgbaColor) {
```
dabbott, line 43: I want to play around with this a bit you rebase, cause I added an interpolate util a couple days ago that we might be able to use to simplify this

packages/noya-colorpicker/src/utils/colorAt.ts
```diff
@@ -0,0 +1,61 @@
...
+        end = stops[i + 1];
+        break;
+      }
+    }
+
+    if (!start || !end) {
+      start = end = stops[stops.length - 1];
+    }
+
+    const step = stepize(start.color, end.color, (end.pos - start.pos) * 100000);
```
dabbott, line 57: What's the purpose of using step sizes? It looks like we're basically making them infinitely small anyway, by using 100000

packages/noya-colorpicker/src/components/Gradient.tsx
```diff
@@ -0,0 +1,91 @@
...
+        onClick={handleClick}
+        isOnPoint={isOnPoint}
+        aria-label="Gradient"
+      >
+        {gradients.map((g, index) => (
+          <Pointer
+            key={`gradients-point-${index}`}
+            selected={index === selectedStop}
+            left={g.position}
+            onClick={() => {
```
dabbott, line 81: I'm wondering if it'd be better to handle all clicks at the `Interactive` level? Cause the user should be able to start dragging immediately after clicking, so I think we may need to handle the click and move events in the same place

packages/noya-colorpicker/src/components/Gradient.tsx
```diff
@@ -0,0 +1,91 @@
...
+              if (moving || index === selectedStop) return;
+              onSelectStop(index);
+              onChange(rgbaToHsva(sketchColorToRgba(g.color)));
```
dabbott, line 84: Why do we change the color when a pointer is clicked?

dabbott (changes requested):
This is looking great so far 😄

Can you merge/rebase with the main branch, so we pick up the new changes?

I noticed that we're still using hover to select gradient stops. We should try to use a click instead.

I'm probably going to have another round of feedback after this.

packages/noya-colorpicker/src/components/Gradient.tsx
milg15: I think it copied this from somewhere 🤔

packages/noya-designsystem/src/utils/getGradientBackground.tsx
milg15: I use this component for showing the preview in the button too, in theory it would get the degree from the gradient to show the direction correctly. If it's undefined it means that we are just showing the gradient bar

packages/noya-colorpicker/src/components/Gradient.tsx
milg15: Good point!

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
```diff
@@ -145,49 +168,166 @@ export default memo(function ColorInputFieldWithPicker({
   id,
   value,
   onChange,
+  onChangeType,
+  onChangeGradientColor,
+  onChangeGradientPosition,
+  onAddGradientStop,
+  onChangeGradientType,
```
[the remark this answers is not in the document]
milg15: The set Layout it's out of this part 🤔 i can't move those buttons to this components.

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
```diff
@@ -145,49 +168,166 @@ export default memo(function ColorInputFieldWithPicker({
...
-  const [state] = useApplicationState();
 
+  const [state, dispatch] = useApplicationState();
   const [swatchLayout, setSwatchLayout] = useState<SwatchLayout>('grid');
 
+  const values = useMemo(() => {
+    if (value._class === 'pattern')
+      //change to show preview of gradient :thinking_emoji:
+      return [];
+    return [value._class === 'gradient' ? value.stops[0].color : value];
```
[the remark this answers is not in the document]
milg15: I had this for the swatch selection 🤔 but if I want to do this I would have to move the gradient index selector to this components. Not sure on this one

packages/noya-state/src/reducers/styleReducer.ts
milg15: Yeah! I was thinking that too

packages/noya-colorpicker/src/components/GradientSlider.tsx
milg15: I'll try combining them and doing the new onClick event 🤔 Maybe they can just work together

packages/noya-colorpicker/src/utils/colorAt.ts
milg15: This is how the logic of the interpolation was 😅 but since you mentioned you had a code for this I think I can use what you are working with

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
milg15: 👍

milg15: I did it but the ContentComponents turns out too big and with a lot of props 🤔  I do think I can separate them in some way but I think at least keeping the fillOption outside would simplify this components. This fillOptions Logic it's used by all the FillType even patterns

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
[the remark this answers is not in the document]
dabbott: Can we move this whole section into a separate component?

Screen Shot 2021-06-14 at 8 52 12 AM

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
dabbott: 👍  I see what you mean. Maybe we leave it like this for now, and revisit when we add "Gradient Presets"

Screenshot_6_14_21__8_57_AM

If we can separate out the swatch stuff now, I think that'll simplify this component a lot already

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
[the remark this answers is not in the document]
dabbott: I actually hadn't thought of this before 😅. In Sketch, there's no way to use a color swatch within a gradient - they don't show the UI for it at all. They instead show "gradient presets", which are like swatches but for gradients.

Screenshot_6_14_21__8_57_AM

We can come back to gradient presets, so maybe for now if the fill type is a gradient, we don't show swatches at all?

packages/noya-designsystem/src/utils/getGradientBackground.tsx
dabbott: Ah ok. I think it'd be clearer to pass the direction explicitly. E.g. `getGradientBackground(values, Sketch.GradientType.Linear, 'horizontal') or something

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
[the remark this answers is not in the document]
milg15: Okey!  I'll just don't show anything in that part for now

packages/app/src/components/inspector/ColorInputFieldWithPicker.tsx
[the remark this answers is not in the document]
milg15: I also added the button to create and detach to this component, Maybe when we add the Gradient Presets I could make this component more generic 🤔 I'll see

dabbott (changes requested):
I merged with main and updated the `colorAt` to use our new interpolate util

packages/noya-colorpicker/src/components/Interactive.tsx
```diff
@@ -97,19 +99,35 @@ const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
...
+          event.target instanceof HTMLElement &&
+          !event.target.classList.contains('pointer')
+        ) {
+          onClickCallback(getRelativePosition(container.current!, event));
+          return;
+        }
+
+        if (event.target instanceof HTMLElement) {
+          //is this correct?
+          event.target.click();
```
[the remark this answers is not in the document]
milg15: This handles the selection of the pointer 🤔 I wasn't sure if I should send the function or just having the function on the pointer works 🤔 Maybe it's a bad practice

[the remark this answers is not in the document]
dabbott: It'd be better to have a prop here, since it's not clear how this works just by letting at the code

packages/noya-colorpicker/src/components/Pointer.tsx
```diff
@@ -26,5 +35,12 @@ export default memo(function Pointer({ left, top = 0.5 }: Props): JSX.Element {
     [left, top],
   );
 
-  return <Container style={style} />;
+  return (
+    <Container
+      className={`pointer ${left}`}
```
dabbott, line 40: Do we need `left` in the classname?

milg15: before but not now 👍

packages/noya-colorpicker/src/components/Gradient.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const handleClick = (interaction: Interaction | number) => {
+    if (typeof interaction === 'number') {
+      onSelectStop(interaction);
+      return;
+    }
+
+    const gradient = gradients.map((g, index) => ({
+      color: sketchColorToRgba(g.color),
+      pos: index === 0 ? 0 : index === 1 ? 1 : g.position,
```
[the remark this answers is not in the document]
milg15: it doesn't need to be 🤔 I think this is not there anymore

packages/noya-colorpicker/src/components/Interactive.tsx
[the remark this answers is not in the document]
milg15: ohh! Okey I was trying to do something like that, so yeah I'll do that instead then

dabbott (approved):
🎉🎉🎉

packages/noya-colorpicker/src/components/Gradient.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const handleClick = (interaction: Interaction | number) => {
+    if (typeof interaction === 'number') {
+      onSelectStop(interaction);
+      return;
+    }
+
+    const gradient = gradients.map((g, index) => ({
+      color: sketchColorToRgba(g.color),
+      position: index === 0 ? 0 : index === 1 ? 1 : g.position,
```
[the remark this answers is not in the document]
milg15: ohh I didn't realized where it was

milg15:
Done! 👍

PR #111, opened 2021-06-09
adding control points

### LINKED ISSUE #107: Render control points
In point editing mode, we should render control points for points with curve mode: Mirrored, Asymmetric, Disconnected.

image

### Implementation

This will probably be entirely in `EditablePath.tsx`.

https://github.com/noya-app/noya/blob/f7fd4d77f7b90eab3a3ebf7ae730d95b7515b428/packages/noya-renderer/src/components/EditablePath.tsx#L88

We can start by updating this `.map` to return the parsedCurvePoints, then create 2 `.map`'s where we render canvas elements: one for `curveFrom` and one for `curveTo`.

Note: we have no way to select a control point yet, so we don't need to consider the selected state.

leannepepper:
Rendering control points 

closes #107

### DIFF
--- a/packages/noya-renderer/src/components/EditablePath.tsx
+++ b/packages/noya-renderer/src/components/EditablePath.tsx
@@ -51,6 +51,47 @@ function EditablePathPoint({ point, fill, stroke }: EditablePathPointProps) {
   );
 }
 
+interface EditableControlPointProps {
+  point: Point;
+  controlPoint: Point;
+  fill: Paint;
+  stroke: Paint;
+}
+
+function ControlPathPoint({
+  point,
+  controlPoint,
+  fill,
+  stroke,
+}: EditableControlPointProps) {
+  const { CanvasKit } = useReactCanvasKit();
+
+  const path = useMemo(() => {
+    const path = new CanvasKit.Path();
+
+    path.addRect(
+      CanvasKit.XYWHRect(
+        controlPoint.x - POINT_SIZE / 2 + 2,
+        controlPoint.y - POINT_SIZE / 2 + 2,
+        POINT_SIZE / 2,
+        POINT_SIZE / 2,
+      ),
+    );
+
+    return path;
+  }, [CanvasKit, controlPoint.x, controlPoint.y]);
+  useDeletable(path);
+  const points = [point, controlPoint];
+
+  return (
+    <>
+      <Polyline points={points} paint={stroke} />
+      <Path path={path} paint={fill} />
+      <Path path={path} paint={stroke} />
+    </>
+  );
+}
+
 interface Props {
   layer: Sketch.AnyLayer;
   transform: AffineTransform;
@@ -85,13 +126,16 @@ export default function EditablePath({
 
   if (!Layers.isPointsLayer(layer)) return null;
 
-  const points = layer.points.map(
-    (point) => Primitives.parseCurvePoint(point, layer.frame).point,
+  const parsedCurvePoints = layer.points.map((point) =>
+    Primitives.parseCurvePoint(point, layer.frame),
   );
+
+  const points = parsedCurvePoints.map((point) => point.point);
 
   return (
     <Group transform={localTransform}>
       <Polyline points={points} paint={stroke} />
+
       {points.map((point, index) => {
         const isSelected = selectedIndexes.includes(index);
 
@@ -101,6 +145,36 @@ export default function EditablePath({
             point={point}
             fill={isSelected ? selectedFill : fill}
             stroke={isSelected ? selectedStroke : stroke}
+          />
+        );
+      })}
+
+      {parsedCurvePoints.map((points, index) => {
+        if (points.curveMode === 1) {
+          return null;
+        }
+        return (
+          <ControlPathPoint
+            key={index}
+            point={points.point}
+            controlPoint={points.curveFrom}
+            fill={fill}
+            stroke={stroke}
+          />
+        );
+      })}
+
+      {parsedCurvePoints.map((points, index) => {
+        if (points.curveMode === 1) {
+          return null;
+        }
+        return (
+          <ControlPathPoint
+            key={index}
+            point={points.point}
+            controlPoint={points.curveTo}
+            fill={fill}
+            stroke={stroke}
           />
         );
       })}
--- a/packages/noya-renderer/src/components/EditablePath.tsx
+++ b/packages/noya-renderer/src/components/EditablePath.tsx
@@ -17,6 +17,7 @@ import React, { Fragment, useMemo } from 'react';
 import { useTheme } from 'styled-components';
 
 const POINT_RADIUS = 4;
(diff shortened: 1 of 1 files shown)


packages/noya-renderer/src/components/EditablePath.tsx
```diff
@@ -104,6 +148,36 @@ export default function EditablePath({
           />
         );
       })}
+
+      {parsedCurvePoints.map((points, index) => {
```
dabbott, line 152: We're currently filtering `parsedCurvePoints` for `curveMode === 1` twice. Let's do a `.filter` before the `return` so we can share this logic

E.g. `const controlPoints = parsedCurvePoints.filter(...)`

packages/noya-renderer/src/components/EditablePath.tsx
```diff
@@ -104,6 +148,36 @@ export default function EditablePath({
...
+            key={index}
+            point={points.point}
+            controlPoint={points.curveFrom}
+            fill={fill}
+            stroke={stroke}
+          />
+        );
+      })}
+
+      {parsedCurvePoints.map((points, index) => {
```
dabbott, line 167: These lines currently render over the purple "point" oval. We should render them beneath instead, by moving them up above `{points.map(...)}`

Screen Shot 2021-06-09 at 8 03 10 PM

packages/noya-renderer/src/components/EditablePath.tsx
```diff
@@ -51,6 +51,47 @@ function EditablePathPoint({ point, fill, stroke }: EditablePathPointProps) {
...
+  stroke,
+}: EditableControlPointProps) {
+  const { CanvasKit } = useReactCanvasKit();
+
+  const path = useMemo(() => {
+    const path = new CanvasKit.Path();
+
+    path.addRect(
+      CanvasKit.XYWHRect(
+        controlPoint.x - POINT_SIZE / 2 + 2,
```
dabbott, line 74: What does the `+ 2` do here?

packages/noya-renderer/src/components/EditablePath.tsx
```diff
@@ -51,6 +51,47 @@ function EditablePathPoint({ point, fill, stroke }: EditablePathPointProps) {
...
+
+    return path;
+  }, [CanvasKit, controlPoint.x, controlPoint.y]);
+  useDeletable(path);
+  const points = [point, controlPoint];
+
+  return (
+    <>
+      <Polyline points={points} paint={stroke} />
+      <Path path={path} paint={fill} />
```
dabbott, line 89: We should use `<Rect>` instead of `<Path>` here, both because it's simpler, and so that we can avoid instantiating a `CanvasKit.Path`

packages/noya-renderer/src/components/EditablePath.tsx
```diff
@@ -51,6 +51,47 @@ function EditablePathPoint({ point, fill, stroke }: EditablePathPointProps) {
...
 }
 
+interface EditableControlPointProps {
+  point: Point;
+  controlPoint: Point;
+  fill: Paint;
+  stroke: Paint;
+}
+
+function ControlPathPoint({
```
dabbott, line 61: This component is rendering 2 things, the control point and the connecting line, but I think it'd be clearer with just 1. Can we try moving the `Polyline` into the parent component, `EditablePath`?

E.g. within the 2 `.maps`:

```tsx
<Fragment key={index}>
  <Polyline
    points={[point.point, point.curveFrom]}
    paint={stroke}
  />
  <ControlPathPoint ... />
</Fragment>
```

packages/noya-renderer/src/components/EditablePath.tsx
```diff
@@ -85,13 +126,16 @@ export default function EditablePath({
 
   if (!Layers.isPointsLayer(layer)) return null;
 
-  const points = layer.points.map(
-    (point) => Primitives.parseCurvePoint(point, layer.frame).point,
+  const parsedCurvePoints = layer.points.map((point) =>
+    Primitives.parseCurvePoint(point, layer.frame),
```
dabbott, line 130: Remember to merge/rebase on main, since we renamed this to `decodeCurvePoint`, and also changed the `POINT_SIZE` => `POINT_RADIUS` above

packages/noya-renderer/src/components/EditablePath.tsx
dabbott: Within these `.map`, we should rename `points` => `point`, since plural implies an array

packages/noya-renderer/src/components/EditablePath.tsx
dabbott: Also probably rename `ControlPathPoint` => `EditablePathControlPoint` and `EditableControlPointProps` => `EditablePathControlPointProps`, for consistency in this file

dabbott (changes requested):
This looks great! Thanks to the smaller PR we can really focus on the details

packages/noya-renderer/src/components/EditablePath.tsx
leannepepper: Done.

packages/noya-renderer/src/components/EditablePath.tsx
leannepepper: Done.

packages/noya-renderer/src/components/EditablePath.tsx
leannepepper: Moves the control point to center of the line. Without it, it looks like this: 
Open to alternative solutions :) 

Screen Shot 2021-06-09 at 8 37 09 PM

packages/noya-renderer/src/components/EditablePath.tsx
leannepepper: Done.

packages/noya-renderer/src/components/EditablePath.tsx
leannepepper: Done.

packages/noya-renderer/src/components/EditablePath.tsx
dabbott: If we make another constant for this, we can use the same formula that we do with the main point:

```ts
const CONTROL_POINT_SIZE = 2

const rect = Primitives.rect(CanvasKit, {
  x: point.x - CONTROL_POINT_SIZE,
  y: point.y - CONTROL_POINT_SIZE,
  width: CONTROL_POINT_SIZE * 2,
  height: CONTROL_POINT_SIZE * 2,
});
```

^ Since the width/height are equal to size * 2, we can offset it by 50% by subtracting half the width/height from x/y

dabbott (approved):

packages/noya-renderer/src/components/EditablePath.tsx
leannepepper: oh good idea, constant has been added.

PR #114, opened 2021-06-13
Click to select point

### LINKED ISSUE #108: Click to select a point
When in point editing mode, the user should be able to click a point to select it.

The user may select multiple points at once. Clicking a point while the shift or cmd key is pressed should add that point to the selection, or remove that point from the selection if it's already selected.

### Implementation

Clicking a point should update the `selectedPointLists`. See: 

https://github.com/noya-app/noya/blob/f7fd4d77f7b90eab3a3ebf7ae730d95b7515b428/packages/noya-state/src/reducers/applicationReducer.ts#L47

https://github.com/noya-app/noya/blob/f7fd4d77f7b90eab3a3ebf7ae730d95b7515b428/packages/noya-state/src/reducers/canvasReducer.ts#L155

We can find point locations by transforming them relative to the canvas's transform and the layer frame, similar to: 

https://github.com/noya-app/noya/blob/f7fd4d77f7b90eab3a3ebf7ae730d95b7515b428/packages/noya-state/src/reducers/pointReducer.ts#L75

Note: we don't need to handle control points as part of this issue. We'll do it next, in a separate issue.

leannepepper:
Adds ability to select points in editPath mode

Closes #108


packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -17,7 +17,22 @@ import { ApplicationState, SetNumberMode } from './applicationReducer';
 export type PointAction =
   | [type: 'setPointCurveMode', curveMode: Sketch.CurveMode]
   | [type: 'setPointCornerRadius', amount: number, mode?: SetNumberMode]
-  | [type: 'setPointX' | 'setPointY', amount: number, mode?: SetNumberMode];
+  | [type: 'setPointX' | 'setPointY', amount: number, mode?: SetNumberMode]
+  | [type: 'selectPoint', point: Point, mode: SetNumberMode];
+
+const POINT_RADIUS = 4;
+
+function isPointInRange(point: Point, rawPoint: Point): boolean {
```
leannepepper, line 25: Should this function be moved to a utility file?

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -41,6 +56,54 @@ export function pointReducer(
...
+          const layer = Layers.access(page, indexPath);
+          const pointList = draft.selectedPointLists[layer.do_objectID];
+          const boundingRect = boundingRects[layer.do_objectID];
+          if (!Layers.isPointsLayer(layer)) return;
+
+          layer.points.forEach((point, index) => {
+            const decodedPoint = decodeCurvePoint(point, boundingRect);
+            if (isPointInRange(decodedPoint.point, rawPoint) || !pointList) {
+              if (pointList.indexOf(index) > -1) {
+                mode === 'replace'
```
leannepepper, line 85: Would these ternary expressions read better if they were switch statements?

packages/app/src/containers/Canvas.tsx
```diff
@@ -249,6 +249,13 @@ export default memo(function Canvas() {
           event.preventDefault();
           break;
         }
+        case 'editPath': {
```
dabbott, line 252: We should probably pass the layerId + index of the target point to `"selectPoint"`, rather than its coordinates. That'll let us re-use the action elsewhere more easily, where we might not have a mouse coordinate. We also want to be more precise once we're in the reducer, since a coordinate might correspond to multiple layer points. And it'll be consistent with how we handle other selections (e.g. `"selectLayer"`). 

For the types, instead of `SetNumberMode`, we should use `SelectionType` like `"selectLayer"` uses:

https://github.com/noya-app/noya/blob/5ffef92f08def4aff8d6fd4e3cdd1dd2e9d01f2c/packages/noya-state/src/reducers/layerReducer.ts#L38-L42

This already has a helper function `updateSelection` for modifying selection arrays. We can call this for each layer in the `selectedPointLists`:

https://github.com/noya-app/noya/blob/5ffef92f08def4aff8d6fd4e3cdd1dd2e9d01f2c/packages/noya-state/src/reducers/layerReducer.ts#L264

For the selection logic, I think we can follow a similar pattern to selecting layers:

https://github.com/noya-app/noya/blob/5ffef92f08def4aff8d6fd4e3cdd1dd2e9d01f2c/packages/app/src/containers/Canvas.tsx#L261-L279

So supposing we made a new type that uniquely identifies a selected point:

```ts
type SelectedPoint = [layerId: string, index: number]
```

Then we would first figure out which point is selected, moving the bounding rect stuff that's in the reducer into `Canvas.tsx`, and afterward we could dispatch `"selectPoint"` actions something like this:

```ts
// Our selected point if we clicked one, or undefined
const selectedPoint: SelectedPoint | undefined = [layer.do_objectID, index]

if (selectedPoint) {
  // Check if the point is already in the selection array
  if (draft.selectedPointLists[layer.do_objectID]?.includes(index)) {
    // If shift is held and more than 1 point is selected
    if (event.shiftKey && Object.values(draft.selectedPointLists).flat().length !== 1) {
      dispatch('selectPoint', selectedPoint, 'difference');
    }
  } else {
    dispatch('selectPoint', selectedPoint, event.shiftKey ? 'intersection' : 'replace');
  }
} else {
  dispatch('selectPoint', undefined);
}
```

^ I tried to adapt the `"selectLayer"` logic so that the behavior is the same

packages/noya-state/src/reducers/pointReducer.ts
dabbott: Yes! We shouldn't use assignment expressions when returning. Looks like there's an eslint rule I can add for this: https://eslint.org/docs/rules/no-return-assign

packages/noya-state/src/reducers/pointReducer.ts
dabbott: Maybe it can go in `pointSelectors`? The most important thing is that we share that `POINT_RADIUS` constant with the rendering code, so that they don't get out of sync

dabbott (changes requested):
Awesome! I think the first step is always just to get it working. Now we can refactor for consistency.

packages/noya-state/src/reducers/pointReducer.ts
leannepepper: isPointInRange and POINT_RADIUS moved to pointSelectors. Constant is now used in` EditablePath` for consistency

packages/noya-state/src/reducers/pointReducer.ts
leannepepper: The more eslint rules the better! Let's add it!

leannepepper: Outer ternary expressions refactored to switch statements.

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -152,7 +152,10 @@ export function canvasReducer(
 
             if (!Layers.isPointsLayer(layer)) break;
 
-            draft.selectedPointLists = { [layer.do_objectID]: [0] };
+            //Selects the first point in the first selected layer
+            layerIds.forEach(function (layer, index) {
```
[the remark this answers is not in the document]
leannepepper: The pervious version only added the first layer to the `selectedPointLists`. So if you had multiple layers selected (copying Sketch behavior here) you could not selected a point that was not part of the first layer.

packages/app/src/containers/Canvas.tsx
leannepepper: Type `SetNumberMode` changed to `SelectionType` in pointReducer. "adjust" changed to "difference" (I'm not sure if that correct)

packages/noya-state/src/reducers/canvasReducer.ts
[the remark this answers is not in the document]
dabbott: I think this new logic is good. 

- We should probably loop over `layerIndexPaths`, rather than `layerIds`, cause we still want to filter out layers that don't have points (e.g. Text layers). I.e. the line above the `if (!Layers.isPointsLayer(layer)) break;` needs to happen for each layer, not just the first.
- I think we can update the comment to say something about initializing a point list for each selected layer

packages/app/src/containers/Canvas.tsx
```diff
@@ -26,6 +30,8 @@ import { useSize } from '../hooks/useSize';
 import { useWorkspace } from '../hooks/useWorkspace';
 import * as MouseEvent from '../utils/mouseEvent';
 
+export type SelectedPoint = [layerId: string, index: number];
```
dabbott, line 33: Let's move this type declaration into the pointsReducer

packages/app/src/containers/Canvas.tsx
```diff
@@ -249,6 +255,68 @@ export default memo(function Canvas() {
           event.preventDefault();
           break;
         }
+        case 'editPath': {
+          const layer: any = Selectors.getLayerAtPoint(
```
dabbott, line 259: `getLayerAtPoint` will affect path selection in a way we don't want, since the POINT_RADIUS extends past the layer's edge. This is probably what's making it harder to click

packages/app/src/containers/Canvas.tsx
```diff
@@ -249,6 +255,68 @@ export default memo(function Canvas() {
...
+              includeHiddenLayers: false,
+            },
+          );
+
+          if (!layer) return;
+
+          const boundingRect = boundingRects[layer.do_objectID];
+          //Why dose TS not recognize layer.points?
+          let selectedPoint: SelectedPoint | undefined = undefined;
+          layer.points.forEach((layer_point: any, index: number) => {
```
dabbott, line 285: Not all layers have points. First we need to filter down the selected layers. Maybe try:

```ts
getSelectedLayers(state).filter(Layers.isPointsLayer).forEach((layer) => { ... }
```

This way we have access to the `layer` and can avoid `getLayerAtPoint`

packages/app/src/containers/Canvas.tsx
```diff
@@ -249,6 +255,68 @@ export default memo(function Canvas() {
...
+            const decodedPoint = decodeCurvePoint(layer_point, boundingRect);
+            if (isPointInRange(decodedPoint.point, point)) {
+              selectedPoint = [layer.do_objectID, index];
+            }
+          });
+
+          if (selectedPoint) {
```
dabbott, line 292: I didn't realize Sketch allows deselecting the only point while holding shift (unlike for layers), so we can simplify as:

```ts
if (selectedPoint) {
  const alreadySelected = state.selectedPointLists[
    selectedPoint[0]
  ]?.includes(selectedPoint[1]);

  dispatch(
    'selectPoint',
    selectedPoint,
    event.shiftKey
      ? alreadySelected
        ? 'difference'
        : 'intersection'
      : 'replace',
  );
} else {
  dispatch('selectPoint', undefined);
}
```

Since I mostly wrote this bit I don't feel bad about rewriting it 😂

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -148,11 +148,19 @@ export function canvasReducer(
 
         switch (interactionState.type) {
           case 'editPath': {
-            const layer = Layers.access(page, layerIndexPaths[0]);
+            const selectedLayers: string[] = [];
 
-            if (!Layers.isPointsLayer(layer)) break;
+            layerIndexPaths.forEach((layerIndex) => {
```
dabbott, line 153: Can we use `flatMap` to make this more functional?

packages/noya-state/src/selectors/pointSelectors.ts
```diff
@@ -7,6 +8,20 @@ import { visit } from '../layers';
 import { getCurrentPage } from './pageSelectors';
 import { getBoundingRectMap } from './selectors';
 
+export const POINT_RADIUS = 4;
+
+export const isPointInRange = (point: Point, rawPoint: Point): boolean => {
+  if (
```
dabbott, line 14: We can simplify by using `return` directly instead of `if/else` here: `return (point.x >= ......)`

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -41,6 +48,23 @@ export function pointReducer(
         curvePoint.cornerRadius = Math.max(0, newValue);
       });
     }
+    case 'selectPoint': {
+      const [, selectedPoint, selectionType] = action;
```
dabbott, line 52: For consistency with other selection actions, and to simplify the loop, lets assign a default value to `selectionType` here: `selectionType = 'replace'`

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -41,6 +48,23 @@ export function pointReducer(
...
+
+      return produce(state, (draft) => {
+        for (let layerId in draft.selectedPointLists) {
+          const currentIds = draft.selectedPointLists[layerId];
+
+          if (!selectionType || !selectedPoint) return;
```
dabbott, line 58: I think we can simplify these conditionals as:

```ts
updateSelection(
  currentIds,
  selectedPoint && selectedPoint[0] === layerId
    ? selectedPoint[1]
    : undefined,
  selectionType,
);
```

packages/noya-state/src/selectors/pointSelectors.ts
```diff
@@ -7,6 +8,20 @@ import { visit } from '../layers';
 import { getCurrentPage } from './pageSelectors';
 import { getBoundingRectMap } from './selectors';
 
+export const POINT_RADIUS = 4;
+
+export const isPointInRange = (point: Point, rawPoint: Point): boolean => {
```
dabbott, line 13: It's probably not noticeable at this size, but wouldn't this select a square instead of a circle? Would this be more accurate?

```ts
distance(point, rawPoint) < POINT_RADIUS
```

dabbott (changes requested):

packages/app/src/containers/Canvas.tsx
leannepepper: replaced with `getSelectedLayers()`

packages/app/src/containers/Canvas.tsx
```diff
@@ -249,6 +261,51 @@ export default memo(function Canvas() {
...
+              clickThroughGroups: true,
+              includeArtboardLayers: false,
+              includeHiddenLayers: false,
+            },
+          );
+          getSelectedLayers(state)
+            .filter(Layers.isPointsLayer)
+            .forEach((layer) => {
+              const boundingRect = boundingRects[layer.do_objectID];
+              layer.points.forEach((layer_point, index) => {
```
dabbott, line 259: Camelcase, maybe `layer_point` => `curvePoint`

dabbott (approved):
Looks great! I found a thing to rename, but I'll go ahead and approve now, so you can merge after

PR #119, opened 2021-06-19
Theme Gradient inspector

milg15:
Allows to change a gradient from a list of gradient and add a new one from the inspector.

image

 _I am still figuring out if there it's a way of attaching the gradient to the current selected gradient and how to delete an existing asset🤔_

### DIFF
--- a/packages/app/src/components/inspector/ColorPickerAssetGrid.tsx
+++ b/packages/app/src/components/inspector/ColorPickerAssetGrid.tsx
@@ -1,0 +1,68 @@
+import { GridIcon, RowsIcon } from '@radix-ui/react-icons';
+import { RadioGroup } from 'noya-designsystem';
+import styled from 'styled-components';
+
+export const PaddedSection = styled.section({
+  padding: '8px 10px',
+  display: 'flex',
+  flexDirection: 'column',
+});
+
+export const Square = styled.div<{ background: string; selected?: boolean }>(
+  ({ theme, background, selected = false }) => ({
+    height: '25px',
+    width: '25px',
+    background,
+    border: `2px solid ${
+      selected ? 'rgb(132,63,255)' : theme.colors.popover.background
+    } `,
+    borderRadius: '4px',
+    cursor: 'pointer',
+  }),
+);
+
+export const GridSmall = styled.div({
+  display: 'grid',
+  gridTemplateColumns: 'repeat(auto-fill, 25px)',
+  gap: '5px',
+});
+
+export const Row = styled.div(({ theme }) => ({
+  flex: '1',
+  display: 'flex',
+  flexDirection: 'row',
+  alignItems: 'stretch',
+}));
+
+export const RadioGroupContainer = styled.div({
+  flex: '0 0 50px',
+  display: 'flex',
+  alignItems: 'stretch',
+});
+
+export type LayoutType = 'list' | 'grid';
+
+export const LayoutPicker = ({
+  layout,
+  setLayout,
+}: {
+  layout: LayoutType;
+  setLayout: (value: LayoutType) => void;
+}) => (
+  <RadioGroupContainer>
+    <RadioGroup.Root
+      id={'layout'}
+      value={layout}
+      onValueChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+        setLayout(event.target.value as LayoutType)
+      }
+    >
+      <RadioGroup.Item value="grid" tooltip="Grid">
+        <GridIcon />
+      </RadioGroup.Item>
+      <RadioGroup.Item value="list" tooltip="List">
+        <RowsIcon />
+      </RadioGroup.Item>
+    </RadioGroup.Root>
+  </RadioGroupContainer>
+);
--- a/packages/app/src/components/inspector/ColorPickerAssetGrid.tsx
+++ b/packages/app/src/components/inspector/ColorPickerAssetGrid.tsx
@@ -1,6 +1,7 @@
 import { GridIcon, RowsIcon } from '@radix-ui/react-icons';
 import { RadioGroup } from 'noya-designsystem';
 import styled from 'styled-components';
+import { memo, useCallback } from 'react';
 
 export const PaddedSection = styled.section({
   padding: '8px 10px',
@@ -42,27 +43,31 @@ export const RadioGroupContainer = styled.div({
 
 export type LayoutType = 'list' | 'grid';
 
-export const LayoutPicker = ({
-  layout,
-  setLayout,
-}: {
-  layout: LayoutType;
-  setLayout: (value: LayoutType) => void;
-}) => (
-  <RadioGroupContainer>
-    <RadioGroup.Root
-      id={'layout'}
-      value={layout}
-      onValueChange={(event: React.ChangeEvent<HTMLInputElement>) =>
-        setLayout(event.target.value as LayoutType)
-      }
-    >
-      <RadioGroup.Item value="grid" tooltip="Grid">
-        <GridIcon />
-      </RadioGroup.Item>
-      <RadioGroup.Item value="list" tooltip="List">
-        <RowsIcon />
-      </RadioGroup.Item>
-    </RadioGroup.Root>
-  </RadioGroupContainer>
+export const LayoutPicker = memo(
+  ({
+    layout,
+    setLayout,
+  }: {
+    layout: LayoutType;
+    setLayout: (value: LayoutType) => void;
+  }) => (
+    <RadioGroupContainer>
+      <RadioGroup.Root
+        id={'layout'}
+        value={layout}
+        onValueChange={useCallback(
+          (event: React.ChangeEvent<HTMLInputElement>) =>
+            setLayout(event.target.value as LayoutType),
+          [setLayout],
+        )}
+      >
+        <RadioGroup.Item value="grid" tooltip="Grid">
+          <GridIcon />
+        </RadioGroup.Item>
+        <RadioGroup.Item value="list" tooltip="List">
+          <RowsIcon />
+        </RadioGroup.Item>
+      </RadioGroup.Root>
+    </RadioGroupContainer>
+  ),
 );
--- a/packages/app/src/components/inspector/ColorPickerGradient.tsx
+++ b/packages/app/src/components/inspector/ColorPickerGradient.tsx
@@ -55,12 +55,14 @@ type SwatchLayout = 'list' | 'grid';
 interface SwatchesProps {
   selectedSwatchId?: string;
   gradients: Sketch.GradientAsset[];
+  gradientType: Sketch.GradientType;
   onSelectGradientAsset: (gradient: Sketch.Gradient) => void;
 }
 
 const SwatchesList = memo(function SwatchesList({
   selectedSwatchId,
   gradients,
+  gradientType,
   onSelectGradientAsset,
 }: SwatchesProps) {
   return (
@@ -68,7 +70,7 @@ const SwatchesList = memo(function SwatchesList({
       {gradients.map(({ do_objectID, gradient, name }) => {
         const colorString = getGradientBackground(
           gradient.stops,
-          gradient.gradientType,
+          gradientType,
           180,
         );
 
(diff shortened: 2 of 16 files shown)


packages/app/src/components/inspector/ColorPickerGradient.tsx
```diff
@@ -0,0 +1,197 @@
...
+    background: color,
+    border: `2px solid ${
+      selected ? 'rgb(132,63,255)' : theme.colors.popover.background
+    } `,
+    borderRadius: '4px',
+    cursor: 'pointer',
+  }),
+);
+
+const GridSmall = styled.div({
```
dabbott, line 34: Can we share most or all of these components with the Swatches grid?

packages/noya-state/src/reducers/gradientReducer.ts
```diff
@@ -98,6 +103,15 @@ export function gradientReducer(
         draft.stops.splice(index, 1);
       });
     }
+    case 'setFillGradientStops':
```
dabbott, line 106: We probably want to set the entire gradient object here, instead of just the stops. Otherwise there's no way to use the other fields of the gradient asset, like `from` and `to`, which are also important.

I think the behavior you have now is good too though, e.g. if you want to use the same stops for a linear vs. radial gradient. Maybe we can expose both at some point.

packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -43,6 +43,7 @@ export type ThemeAction =
       mode?: SetNumberMode,
     ]
   | [type: `addSwatch`, name?: string, color?: Sketch.Color, id?: string]
+  | [type: `addGradientStyle`, name: string, value: Sketch.Gradient]
```
dabbott, line 46: Maybe `addGradientAsset` or `addGradient`? A "style" should usually refer to a `Sketch.Style`

dabbott (commented):
This looks good so far! 

I don't see anything in the sketch file for associating a gradient asset with the current gradient 🤷‍♂️. We could use `isDeepEqual`, but I don't think it's that important really

Do you want to include the context menu for deleting in this PR, or do it separately?

milg15:
I can add the context menu in this PR. I am figuring out how the contextMenu component works for this particular case I am getting some erros since the selection of the gradient happensa inside the Component not with the AppState🤔

dabbott:
Oh yeah, that's a good point. We might need to move `selectedStopIndex` up into `ColorInputFieldWithPicker`?

milg15:
> Oh yeah, that's a good point. We might need to move `selectedStopIndex` up into `ColorInputFieldWithPicker`?

I have a solution for this, where I can keep the useState inside the Grid/List viewer but the issue itsn't exactly with the selector but the fact that the OnContext needs a event parameter but the list component doesn't require it 🤔

packages/noya-state/src/reducers/gradientReducer.ts
milg15: That's why I wasn't sure about showing the same gradient themes for the linear, radial or angular

packages/app/src/components/inspector/ColorPickerGradient.tsx
milg15: Do I keep this Compoments in SwatchesGrid ?

packages/app/src/components/inspector/ColorPickerGradient.tsx
```diff
@@ -176,21 +206,28 @@ export default memo(function ColorPickerSwatches({
...
+        <ContextMenu.Root<MenuItemType>
+          items={menuItems}
+          onSelect={handleSelectMenuItem}
+        >
+          {gradientLayout === 'grid' ? (
+            <SwatchesGrid
+              gradients={gradientAssets}
+              gradientType={gradientType}
+              setGradientId={setGradientId}
+              onContextMenu={() => {}}
```
milg15, line 218: For some reason if I don't add this, the contextMenu doesn't show at all😅

packages/noya-state/src/reducers/gradientReducer.ts
dabbott: This might already be clear but just in case, we should still show all gradient themes regardless of the current gradient type:

Screen Shot 2021-06-21 at 2 02 19 PM

Picking a gradient asset will switch the gradient type if needed

packages/app/src/components/inspector/ColorPickerGradient.tsx
```diff
@@ -73,15 +52,19 @@ const SwatchesList = memo(function SwatchesList({
           gradientType,
           180,
         );
+        const handleContextMenu = () => {
```
milg15, line 55: If I don't send the event?: React.MouseEvent, to this onContextMenu it doesn't trigger for the list. I need it to look like this so I can set which gradient I am currently selecting

packages/app/src/components/inspector/ColorPickerGradient.tsx
```diff
@@ -121,22 +106,67 @@ const SwatchesGrid = memo(function SwatchesGrid({
...
   gradientAssets,
   onChange,
   onCreate,
+  onRename,
+  onDelete,
 }: Props) {
   const [gradientLayout, setGradientLayout] = useState<SwatchLayout>('grid');
+  const [gradientId, setGradientId] = useState('');
+
+  const handleSelectMenuItem = useCallback(
```
milg15, line 153: am I missing something here ? I am not sure why when I hover on an option the item doesn't highlight

packages/noya-state/src/reducers/gradientReducer.ts
milg15: Ohhhhh okey I understand now

packages/app/src/components/inspector/ColorPickerGradient.tsx
dabbott: I tried it a bit. I think the highlighting issue is a bug in the library that happens when there's a popup open. We don't need to fix this

packages/app/src/components/inspector/ColorPickerGradient.tsx
dabbott: Our `ListView.Row` already has a built-in menu, so we should be able to use `menuItems` and `onSelectMenuItem` to get a context menu. There might still be problems afterward, but lets try that first

packages/app/src/components/inspector/ColorPickerGradient.tsx
dabbott: I'm not sure, but I think we want `<ContextMenu.Root<MenuItemType>` around the individual `<Square />` components. It's a bit counterintuitive but for our ListView and GridView, we have a separate `<ContextMenu.Root<MenuItemType>` for each item, so we should be consistent with that approach. Let's try that first and see if there are still issues

packages/app/src/components/inspector/ColorPickerGradient.tsx
dabbott: Maybe move to a new file `inspector/ColorPickerAssetGrid`

packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -666,6 +668,48 @@ export function themeReducer(
...
+      });
+    }
+    case 'removeGradientAsset': {
+      const [, id] = action;
+      const ids = typeof id === 'string' ? [id] : id;
+
+      return produce(state, (draft) => {
+        const assets = draft.sketch.document.assets;
+
+        assets.gradientAssets = assets.gradientAssets.filter((g, index) => {
```
dabbott, line 695: I think this loop can leave `assets.gradients` in a different state depending on its length and the ids passed, and I'm assuming it's supposed to match `assets.gradientAssets`.

Maybe we can look up the indexes to remove first, and then remove them in a separate pass

```ts
const indexes = assets.gradientAssets.flatMap(
  (g, index) => ids.includes(g.do_objectID) ? [index] : []
)

indexes.reverse().forEach((index) => {
  // splice from both arrays
})
```

packages/app/src/components/inspector/ColorPickerGradients.tsx
```diff
@@ -0,0 +1,208 @@
...
+    <GridSmall>
+      {gradients.map(({ do_objectID, gradient }) => {
+        const gridString = getGradientBackground(
+          gradient.stops,
+          gradient.gradientType,
+          180,
+        );
+
+        return (
+          <ContextMenu.Root<MenuItemType>
```
dabbott, line 89: We should move the `key` from `GridItem` to here on `ContextMenu.Root`

packages/app/src/components/inspector/ColorPickerGradients.tsx
```diff
@@ -0,0 +1,208 @@
...
+    <ListView.Root>
+      {gradients.map(({ do_objectID, gradient, name }) => {
+        const colorString = getGradientBackground(
+          gradient.stops,
+          gradient.gradientType,
+          180,
+        );
+
+        return (
+          <ContextMenu.Root<MenuItemType>
```
dabbott, line 52: We should move the `key` from `ListView.Row` to here on `ContextMenu.Root`

packages/app/src/components/inspector/ColorPickerAssetGrid.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const RadioGroupContainer = styled.div({
+  flex: '0 0 50px',
+  display: 'flex',
+  alignItems: 'stretch',
+});
+
+export type LayoutType = 'list' | 'grid';
+
+export const LayoutPicker = ({
```
dabbott, line 45: We should wrap this in `memo`

packages/app/src/components/inspector/ColorPickerAssetGrid.tsx
```diff
@@ -0,0 +1,68 @@
...
+  setLayout,
+}: {
+  layout: LayoutType;
+  setLayout: (value: LayoutType) => void;
+}) => (
+  <RadioGroupContainer>
+    <RadioGroup.Root
+      id={'layout'}
+      value={layout}
+      onValueChange={(event: React.ChangeEvent<HTMLInputElement>) =>
```
dabbott, line 56: `useCallback`

dabbott (changes requested):
💯 I tried it out and I think it works great! Just a bunch of little things left

packages/app/src/components/inspector/ColorPickerGradients.tsx
```diff
@@ -0,0 +1,208 @@
...
+              onContextMenu={() => setGradientId(do_objectID)}
+            />
+          </ContextMenu.Root>
+        );
+      })}
+    </GridSmall>
+  );
+});
+
+const GridItem = memo(
```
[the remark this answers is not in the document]
milg15: oh I thought you always needed it

packages/app/src/components/inspector/ColorPickerGradients.tsx
```diff
@@ -0,0 +1,208 @@
...
+        case 'delete': {
+          onDelete(gradientId);
+          break;
+        }
+      }
+    },
+    [gradientId, onRename, onDelete],
+  );
+
+  if (!onChange) return null;
```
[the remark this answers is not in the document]
milg15: because the onGradientChange it's optional in the ColorPicker

packages/app/src/components/inspector/GradientInspector.tsx
```diff
@@ -51,6 +51,12 @@ export default memo(function GradientInspector({
 
   const [selectedStopIndex, setSelectedStopIndex] = useState(0);
 
+  /*TODO: Fix error when modifying the lenght of gradientsStops :thinking:
```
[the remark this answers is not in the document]
milg15: It's not resolved

packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -666,6 +668,52 @@ export function themeReducer(
...
+    }
+    case 'setGradientAssetName': {
+      const [, id, name] = action;
+      const ids = typeof id === 'string' ? [id] : id;
+
+      return produce(state, (draft) => {
+        const gradientAssets = draft.sketch.document.assets.gradientAssets;
+
+        gradientAssets.forEach((g) => {
+          if (ids.includes(g.do_objectID)) g.name = name;
```
dabbott, line 713: Do we want to update the name in `assets.gradients` here too? E.g. using the `indexes` approach above

dabbott (approved):
💯

packages/noya-state/src/reducers/themeReducer.ts
milg15: the gradients don't have names, only the gradientAssets

PR #120, opened 2021-06-19
Control point inspector

### LINKED ISSUE #110: Control point inspector
When a control point is selected, we should allow modifying the point's coordinates via the inspector. 

As part of this issue, we should also ensure control points are valid when the type is changed, e.g. from Straight to Mirrored:

Screen Shot 2021-06-09 at 9 59 14 AM

New control points should be created at a distance of 1/3 the length smallest adjacent edge, "tangential" to the shape (e.g. the same slope as if we connected the 2 adjacent points with a line).

### Implementation

We should show the `CoordinatesInspector.tsx` component.

image

We should make a new `ControlPointCoordinatesInspector`, and add new actions `'setControlPointX'` and `'setControlPointY'`.

This will be analogous to our `PointCoordinatesInspector`:

https://github.com/noya-app/noya/blob/f7fd4d77f7b90eab3a3ebf7ae730d95b7515b428/packages/app/src/containers/PointCoordinatesInspector.tsx

There's one bit of additional complexity for control points however: if the type of the point is `Mirrored` or `Asymmetric`, then updating one control point also updates the opposite control point. We should handle this both when changing coordinates and point type.

### Measuring the layer's frame

The layer's frame is the bounding rect containing the hull of the same. When a control point changes, we'll need to determine the layer's frame. We should consider using CanvasKit's `computeTightBounds` to determine this.

leannepepper:
Adding ability to change control points in pathEdit mode. 

closes #110

### DIFF
--- a/packages/app/src/containers/PointControlsInspector.tsx
+++ b/packages/app/src/containers/PointControlsInspector.tsx
@@ -56,7 +56,6 @@ export default memo(function PointControlsInspector() {
   const [state, dispatch] = useApplicationState();
 
   const controlPoint = Selectors.getSelectedControlPoint(state);
-  // eslint-disable-next-line react-hooks/exhaustive-deps
   const points = controlPoint
     ? [controlPoint]
     : Selectors.getSelectedPoints(state);
--- a/packages/app/src/containers/PointControlsInspector.tsx
+++ b/packages/app/src/containers/PointControlsInspector.tsx
@@ -55,7 +55,11 @@ function getCurveModeString(value: Sketch.CurveMode): CurveModeOption {
 export default memo(function PointControlsInspector() {
   const [state, dispatch] = useApplicationState();
 
-  const points = Selectors.getSelectedPoints(state);
+  let points = Selectors.getSelectedPoints(state);
+  if (points.length === 0) {
+    const controlPoint = Selectors.getSelectedControlPoint(state);
+    controlPoint ? points.push(controlPoint) : (points = []);
+  }
 
   const curveMode = useMemo(() => {
     const value = getMultiValue(points.map((point) => point.curveMode));
--- a/packages/app/src/containers/PointControlsInspector.tsx
+++ b/packages/app/src/containers/PointControlsInspector.tsx
@@ -56,13 +56,10 @@ export default memo(function PointControlsInspector() {
   const [state, dispatch] = useApplicationState();
 
   const controlPoint = Selectors.getSelectedControlPoint(state);
-  const points = useMemo(() => {
-    const value = controlPoint
-      ? [controlPoint]
-      : Selectors.getSelectedPoints(state);
-
-    return value;
-  }, [controlPoint, state]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const points = controlPoint
+    ? [controlPoint]
+    : Selectors.getSelectedPoints(state);
 
   const curveMode = useMemo(() => {
     const value = getMultiValue(points.map((point) => point.curveMode));
--- a/packages/app/src/containers/PointControlsInspector.tsx
+++ b/packages/app/src/containers/PointControlsInspector.tsx
@@ -55,11 +55,14 @@ function getCurveModeString(value: Sketch.CurveMode): CurveModeOption {
 export default memo(function PointControlsInspector() {
   const [state, dispatch] = useApplicationState();
 
-  let points = Selectors.getSelectedPoints(state);
-  if (points.length === 0) {
-    const controlPoint = Selectors.getSelectedControlPoint(state);
-    controlPoint ? points.push(controlPoint) : (points = []);
-  }
+  const controlPoint = Selectors.getSelectedControlPoint(state);
+  const points = useMemo(() => {
+    const value = controlPoint
+      ? [controlPoint]
+      : Selectors.getSelectedPoints(state);
+
+    return value;
+  }, [controlPoint, state]);
 
   const curveMode = useMemo(() => {
     const value = getMultiValue(points.map((point) => point.curveMode));
--- a/packages/app/src/containers/PointControlsInspector.tsx
+++ b/packages/app/src/containers/PointControlsInspector.tsx
@@ -7,7 +7,7 @@ import {
   Spacer,
 } from 'noya-designsystem';
 import { Selectors } from 'noya-state';
-import { memo, useCallback, useMemo } from 'react';
+import { memo, useCallback } from 'react';
 import * as InspectorPrimitives from '../components/inspector/InspectorPrimitives';
 import { useApplicationState } from '../contexts/ApplicationStateContext';
 import getMultiNumberValue from '../utils/getMultiNumberValue';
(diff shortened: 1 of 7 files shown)


packages/app/src/containers/PointControlsInspector.tsx
```diff
@@ -55,7 +55,11 @@ function getCurveModeString(value: Sketch.CurveMode): CurveModeOption {
 export default memo(function PointControlsInspector() {
   const [state, dispatch] = useApplicationState();
 
-  const points = Selectors.getSelectedPoints(state);
+  let points = Selectors.getSelectedPoints(state);
+  if (points.length === 0) {
+    const controlPoint = Selectors.getSelectedControlPoint(state);
+    controlPoint ? points.push(controlPoint) : (points = []);
```
dabbott, line 61: Outside of immer's `produce`, we should avoid methods that mutate arrays like `push` when possible. We should also avoid assignments in ternary expressions. Maybe:

```ts
const controlPoint = Selectors.getSelectedControlPoint(state);
const points = controlPoint ? [controlPoint] : Selectors.getSelectedPoints(state)
```

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -172,6 +187,133 @@ export function pointReducer(
...
+            .forEach((curvePoint) => {
+              const decodedPoint = decodeCurvePoint(curvePoint, boundingRect);
+              if (!controlPointType) return state;
+
+              //TODO: Fix Asymmetric point values
+
+              if (
+                (controlPointType === 'curveFrom' ||
+                  controlPointType === 'curveTo') &&
+                (curveMode === 2 || curveMode === 3)
```
dabbott, line 231: We can use enums like `Sketch.CurveMode.Mirrored`, etc, instead of the numbers directly

packages/noya-state/src/selectors/pointSelectors.ts
```diff
@@ -57,3 +57,49 @@ export const getSelectedPoints = (
 
   return points;
 };
+
+export const getSelectedControlPoint = (
```
dabbott, line 61: It looks like this is about the same as `getSelectedPoints`. Can we try sharing the main logic?

For example, a new function that they both call, 

```ts
const getSelectedPointsFromPointLists = (
  state: ApplicationState,
  pointLists: Record<string, number[]>,
): DecodedCurvePoint[] => {
  // ...
};
```

When we have a `selectedControlPoint`, we can make a new "pointsList" with just one point, `{ [layerId]: [pointIndex] }`, and then take just the first result from that function. Which is pretty similar to what you have now.

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -172,6 +187,133 @@ export function pointReducer(
...
+                  curvePoint.curveTo = encodedPoint.curveTo;
+                }
+              }
+            });
+
+          const decodedPoints = layer.points.map((curvePoint) =>
+            decodeCurvePoint(curvePoint, boundingRect),
+          );
+
+          const layerPath = path(
```
dabbott, line 290: Maybe clearer as `const [minX, minY, maxX, maxY] = path(...)`, if that's accurate

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -172,6 +187,133 @@ export function pointReducer(
...
+    case 'setControlPointY': {
+      if (!state.selectedControlPoint) return state;
+      const [type, amount, controlPointType, CanvasKit, mode] = action;
+      const axis = type === 'setControlPointX' ? 'x' : 'y';
+
+      const pageIndex = getCurrentPageIndex(state);
+      const layerIndexPaths = getSelectedLayerIndexPaths(state);
+      const boundingRects = getBoundingRectMap(
+        getCurrentPage(state),
+        [state.selectedControlPoint?.layerId],
```
dabbott, line 200: We don't need `?`

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -172,6 +187,133 @@ export function pointReducer(
...
+          includeArtboardLayers: false,
+          includeHiddenLayers: false,
+        },
+      );
+
+      return produce(state, (draft) => {
+        layerIndexPaths.forEach((indexPath) => {
+          const page = draft.sketch.pages[pageIndex];
+          const layer = Layers.access(page, indexPath);
+          const pointList = [draft.selectedControlPoint?.pointIndex];
```
dabbott, line 212: We don't need to make an array here

dabbott (commented):
Looks great so far! Since it's still in progress, I'll wait till we get the asymmetric points in before leaving more comments on the reducer

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -172,6 +182,168 @@ export function pointReducer(
         });
       });
     }
+    case 'setControlPointX':
+    case 'setControlPointY': {
+      if (!state.selectedControlPoint) return state;
+      const [type, amount, CanvasKit, mode] = action;
+      const controlPointType = state.selectedControlPoint.controlPointType;
```
dabbott, line 189: We should probably destructure the whole `selectedControlPoint` here, since otherwise TypeScript may think they're optional later.

```ts
const { layerId, pointIndex, controlPointType } = state.selectedControlPoint
```

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -172,6 +182,168 @@ export function pointReducer(
...
+          const pointList = draft.selectedControlPoint?.pointIndex;
+          const curveMode = draft.selectedControlPoint?.curveMode;
+          const boundingRect = boundingRects[layer.do_objectID];
+
+          if (!Layers.isPointsLayer(layer) || !boundingRect) return;
+
+          // Update all points by first transforming to the canvas's coordinate system
+          layer.points
+            .filter((_, index) => pointList === index)
+            .forEach((curvePoint) => {
```
dabbott, line 217: Can we make this a `map` instead of a `forEach`? It's usually easier to maintain code that doesn't mutate state, and in this case we have to map the points later to encode them anyway so there's no immer-related perf gain here

dabbott (changes requested):
The core algorithm looks great! There are just a bunch of small things to clean up

packages/app/src/containers/PointControlsInspector.tsx
```diff
@@ -55,7 +55,14 @@ function getCurveModeString(value: Sketch.CurveMode): CurveModeOption {
 export default memo(function PointControlsInspector() {
   const [state, dispatch] = useApplicationState();
 
-  const points = Selectors.getSelectedPoints(state);
+  const controlPoint = Selectors.getSelectedControlPoint(state);
+  const points = useMemo(() => {
```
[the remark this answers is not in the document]
leannepepper: Will do, but I'll have to disable one of the eslint rules for this line. 
`eslint-disable-next-line react-hooks/exhaustive-deps`

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -172,6 +182,168 @@ export function pointReducer(
...
+              const decodedPoint = decodeCurvePoint(curvePoint, boundingRect);
+              if (!controlPointType) return state;
```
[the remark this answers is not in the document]
leannepepper: Yes, when theres point values in state.selectedPointLists, controlPointType will be undefined.

packages/app/src/containers/PointControlsInspector.tsx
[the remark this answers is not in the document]
dabbott: Oh! I just mean we can get rid of `useMemo` here.

Same with the `useMemo` for `curveMode` right below, but it doesn't really matter there.

packages/noya-state/src/reducers/pointReducer.ts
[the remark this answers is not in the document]
dabbott: Ah, I see. I left a comment above about destructuring everything at the top, which I think will solve this

dabbott (approved):

PR #123, opened 2021-06-24
Pattern inspector

milg15:
TODO: 

- [x] Change Image 
- [x] Rename relevant files
- [x] Set "initializer" image


milg15:
There are some errors in the preview but it should be fix on the next issue when we change the previews to use the Canvas instead of css.

packages/app/src/components/inspector/FillInputFieldWithPicker.tsx
```diff
@@ -64,32 +62,29 @@ interface Props {
...
-
-  const values = useMemo(() => {
-    if (value._class !== 'color') return [];
-    return [value];
-  }, [value]);
-
-  const selectedColor = values[0];
+  onChangeFillImage,
+}: PickersProps) => {
+  const [state, dispatch] = appState;
```
dabbott, line 87: Can we call `useApplicationState` here instead of passing these as props?

packages/app/src/components/inspector/GradientInspector.tsx
```diff
@@ -77,36 +73,28 @@ export default memo(function GradientInspector({
...
-      } else {
-        onChangeColor(
-          {
-            ...selectedcolor,
-            alpha: clamp(selectedcolor.alpha + amount / 100, 0, 1),
-          },
-          clampedSelectedStopIndex,
-        );
-      }
+      onChangeColor(
```
dabbott, line 89: I don't think we want to delete the `onNudgeOpacity` case here, since that'll be important when we support modifying fills for multiple layers, all with different colors. In that case, we wouldn't want to override the entire color. Probably same with `onChangeOpacity` above

packages/app/src/components/inspector/PatternInspector.tsx
```diff
@@ -0,0 +1,238 @@
...
+      <Container
+        onDragOver={handleDragEvent}
+        onDragEnter={(e: DragEvent) => handleDragEvent(e, true)}
+        onDragLeave={(e: DragEvent) => handleDragEvent(e, false)}
+        onDrop={handleDropEvent}
+        isDragOver={isDragOver}
+        background={background}
+        backgroundSize={backgroundSize}
+        repeat={isTile}
+        onMouseEnter={() => setShowButton(true)}
```
dabbott, line 198: We have a `useHover` hook which we can use for this that handles some pointer weirdness. See:

https://github.com/noya-app/noya/blob/60d6525289203f0a325f805e7621ab50dc899261/packages/noya-designsystem/src/components/ListView.tsx#L178

We can pass a `onHoverChange` function to this hook to get updates about hover state

packages/app/src/components/inspector/PatternInspector.tsx
```diff
@@ -0,0 +1,238 @@
...
+  const backgroundSize = useMemo(
+    () => getPatternSize(pattern.patternFillType, pattern.patternTileScale),
+    [pattern.patternFillType, pattern.patternTileScale],
+  );
+
+  const handleImageFile = useCallback(
+    async (file: File) => {
+      const data = await file.arrayBuffer();
+
+      const extension = file.type.split('/')[1];
```
dabbott, line 140: We'll probably want a mapping from type => extension, both to show an error for invalid file types, and to handle jpeg. E.g.

```ts
const SUPPORTED_FILE_TYPES = {
  'image/png': 'png',
  'image/jpeg': 'jpg',
  'image/webp': 'webp'
}

// ...

const extension = SUPPORTED_FILE_TYPES[file.type]

if (!extension) {
  alert(`Files of type ${file.type} aren't supported. Files of type png, jpg, and webp are supported.`)
}
```

packages/noya-designsystem/src/utils/sketchPattern.ts
```diff
@@ -0,0 +1,69 @@
+import Sketch from '@sketch-hq/sketch-file-format-ts';
+import { FileMap } from 'noya-sketch-file';
+
+export const imgFileExtensions = ['png', 'jpeg', 'webp'];
+export const mimeTypes = imgFileExtensions.map(
+  (e) => `image/${e === 'ico' ? 'x-icon' : e}`,
```
dabbott, line 6: I don't think we need to handle `ico` or `x-icon` files. Did you encounter this case?

packages/app/src/components/inspector/PatternInspector.tsx
dabbott: If we still need `mimeTypes` or `extensions` as an array, we can do `Object.keys(SUPPORTED_FILE_TYPES)` or `Object.values(SUPPORTED_FILE_TYPES)`

packages/noya-designsystem/src/utils/sketchPattern.ts
```diff
@@ -0,0 +1,69 @@
...
+);
+
+export type SketchPattern = {
+  _class: 'pattern';
+  image?: Sketch.FileRef | Sketch.DataRef;
+  patternFillType: Sketch.PatternFillType;
+  patternTileScale: number;
+};
+
+function encodeImg(input: Uint8Array) {
```
dabbott, line 16: Since we have this in a separate file already, should we delete it here? Presumably we'll end up replacing this with CanvasKit API calls anyway, so not super important

packages/noya-designsystem/src/utils/sketchPattern.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const bytes = new Uint8Array(images[image._ref]);
+  return `url('data:image/png;base64,${encodeImg(bytes)}')`;
+}
+
+export function getPatternSize(
+  fillType: Sketch.PatternFillType,
+  tileScale: number,
+) {
+  return fillType === Sketch.PatternFillType.Fit
```
dabbott, line 62: Maybe a `switch` would be clearer?

packages/noya-state/src/reducers/styleReducer.ts
```diff
@@ -301,6 +319,13 @@ export function styleReducer(
         if (!draft.fills[index].gradient) {
           draft.fills[index].gradient = Models.fill.gradient;
         }
+        if (!draft.fills[index].image) {
```
dabbott, line 322: Maybe we should only do this if the `type` is pattern, just so we don't create objects we don't need in the save data. Same with gradient above

packages/noya-state/src/reducers/styleReducer.ts
```diff
@@ -354,6 +379,53 @@ export function styleReducer(
...
+      const [, index, amount, mode = 'replace'] = action;
+      return produce(state, (draft) => {
+        if (!draft.fills || !draft.fills[index]) return;
+
+        const newValue =
+          mode === 'replace'
+            ? amount
+            : draft.fills[index].patternTileScale + amount;
+
+        draft.fills[index].patternTileScale = Math.min(
```
dabbott, line 400: We can use `clamp` here

packages/noya-state/src/reducers/styleReducer.ts
```diff
@@ -354,6 +379,53 @@ export function styleReducer(
...
+    case 'setFillContextSettingsOpacity': {
+      const [, index, amount, mode = 'replace'] = action;
+      return produce(state, (draft) => {
+        if (!draft.fills || !draft.fills[index]) return;
+        const newValue =
+          mode === 'replace'
+            ? amount
+            : draft.fills[index].contextSettings.opacity + amount;
+
+        draft.fills[index].contextSettings.opacity = Math.min(
```
dabbott, line 423: Also `clamp` here

packages/app/src/components/inspector/PatternInspector.tsx
```diff
@@ -0,0 +1,238 @@
...
+    const file = await fileOpen({
+      extensions: imgFileExtensions.map((e) => `.${e}`),
+      mimeTypes: mimeTypes,
+    });
+
+    handleImageFile(file);
+  }, [handleImageFile]);
+
+  const handleDragEvent = useCallback((e: DragEvent, on?: boolean) => {
+    if (on) setIsDragOver(on);
```
dabbott, line 163: Should this be `if (on !== undefined)`? Otherwise `setIsDragOver` will not be called for `false`

packages/app/src/components/inspector/PatternInspector.tsx
```diff
@@ -0,0 +1,238 @@
...
+    e.preventDefault();
+    // Do something
+  }, []);
+
+  const handleDropEvent = useCallback(
```
dabbott, line 168: Let's try separating the drag and drop handling into a `hooks/useFileDropTarget.ts`.

```ts
function useFileDropTarget() {
  // useState stuff
  return {
    dropTargetProps: {
      onDragOver: ...,
      onDragEnter: ...,
      onDragLeave: ...,
      onDrop: ...
    },
    isDropTargetActive, // maybe instead of `isDropOver`
  }
}
```

Then this component can use it:

```tsx
const { dropTargetProps, isDropTargetActive } = useFileDropTarget()

return (
  // ...
    <Container {...dropTargetProps} isActive={isDropTargetActive}>...</Container>
  // ...
)
```

packages/app/src/components/inspector/FillRow.tsx
```diff
@@ -107,12 +124,24 @@ export default memo(function ColorFillRow({
...
+    (value: PatternFillTypes) => {
+      if (onChangePatternFillType)
+        onChangePatternFillType(Sketch.PatternFillType[value]);
+    },
+    [onChangePatternFillType],
+  );
+
+  /**?
+   *
+   * Join all the onChange ?? :thinking:
```
dabbott, line 137: I think I prefer it like this, since otherwise we'll have to ignore certain types in the callbacks of border/shadow pickers

packages/app/src/components/inspector/PatternInspector.tsx
```diff
@@ -0,0 +1,238 @@
...
+    display: show ? 'block' : 'none',
+    zIndex: 1,
+  }),
+);
+
+interface Props {
+  id: string;
+  images: FileMap;
+  pattern: SketchPattern;
+  onChangeImage?: (image: Sketch.FileRef | Sketch.DataRef) => void;
```
dabbott, line 76: Do these need to be optional? Probably if these callbacks don't exist we shouldn't render this component within the parent at all

dabbott (changes requested):
🔥 This looks great! I found a few things we can try refactoring before we move on to the canvas portion of this.

packages/app/src/components/inspector/PatternInspector.tsx
milg15: this is the same issue I have with gradients. All the actions are options because the are optional in the FillPicker. It's there a way of telling it that if the value._class it's gradient and patterns, those functions exist for sure ? 🤔

packages/noya-designsystem/src/utils/sketchPattern.ts
milg15: If we just leave it to the extensions we have. I'll remove that part  👍

packages/app/src/components/inspector/PatternInspector.tsx
dabbott: I think so... l'll see if I can come up with something tomorrow morning

packages/app/src/components/inspector/PatternInspector.tsx
milg15: ohhhhh!! That explains a lot! right 😆

packages/app/src/components/inspector/PatternInspector.tsx
milg15: I would do something similar to the onChanges. I think that way we won't have that many props in some components

dabbott: I tried improving this, but couldn't come up with anything great. The approach you suggested of using `value._class` look like it isn't supported yet but will be soon https://github.com/microsoft/TypeScript/pull/38839

I did see a few other things I want to try refactoring though, after we merge this

milg15: My brain it's in the future on this one 😆 I'll work on the other changes while I think of a solution for this

packages/noya-designsystem/src/utils/sketchPattern.ts
milg15: delete the encodeImg funcion? I just have it for preview for now. I'll delete it when we change to use CanvasKit

dabbott: Is this the same as what's in `utils/encodeImg.ts`? Or is it different?

milg15: Oh i forgot about that file 😅 I don't think we need it

packages/app/src/components/inspector/GradientInspector.tsx
milg15: 🤔 okey!

packages/app/src/components/inspector/PatternInspector.tsx
milg15: I work on everything but this part

packages/app/src/components/inspector/GradientInspector.tsx
dabbott: We should probably still put the nudge cases back in here and above

packages/app/src/components/inspector/PatternInspector.tsx
```diff
@@ -0,0 +1,242 @@
...
+  flexDirection: 'column',
+}));
+
+const Container = styled.div<{
+  background: string;
+  backgroundSize: string;
+  repeat: boolean;
+  isActive: boolean;
+}>(({ theme, background, backgroundSize, repeat, isActive }) => {
+  const dragOverlayColor =
```
dabbott, line 33: This is clever, but will break if we change the color to a hex code or something. Let's add a new theme color for this overlay color

packages/app/src/components/inspector/PatternInspector.tsx
```diff
@@ -0,0 +1,242 @@
...
+      mimeTypes: Object.keys(SUPPORTED_FILE_TYPES),
+    });
+
+    handleImageFile(file);
+  }, [handleImageFile]);
+
+  const handleDropEvent = useCallback(
+    async (e: DragEvent) => {
+      e.preventDefault();
+      for (const file of e.dataTransfer.items) {
```
dabbott, line 175: This looks like it will call `handleImageFile` for each file in a list, but we probably only want to handle one. We could do:

```ts
const file = e.dataTransfer.items[0]

if (!file) return 
      
const entry = ...;
```

We should probably also show some user feedback with `alert` if the `kind` is wrong or we can't get the file as an `entry`

dabbott (approved):
Now it's working for me! I dunno what it was... anyway, looks great!

packages/app/src/hooks/useFileDropTarget.ts
```diff
@@ -0,0 +1,28 @@
+import { useCallback, useState, DragEvent } from 'react';
+
+export function useFileDropTarget(dropEvent: (e: DragEvent) => Promise<void>) {
```
dabbott, line 3: Can `dropEvent` return just `void` here?

packages/app/src/components/inspector/PatternInspector.tsx
milg15: What should it say when we have the wrong kind or entry?

dabbott: - For wrong kind, "Only files are supported"
- For invalid entry, "Unable to read file ${filename}"

milg15: We don't use kind anymore, I change the to files instead of items. Because items it's used mostly when you want to also allow to drop directories and I don't think we are interested in that.

milg15: I can't get the file name if the file it's invalid 😅

milg15:
I think with those changes it should be good to go 👍

PR #126, opened 2021-06-25
Export preview

dabbott:
We now show an export preview image for every kind of layer. We also have a basic way to export a layer as a single PNG.

TODO:
- Exporting using the defined export formats
- Fix an issue where the size of the inspector jitters due to the preview's automatic resizing (fix with custom scroll area)

Screen Shot 2021-06-25 at 1 24 26 PM
Screen Shot 2021-06-25 at 1 24 48 PM
Screen Shot 2021-06-25 at 1 24 07 PM

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -6,6 +6,7 @@
     "@dnd-kit/core": "^1.0.1",
     "@dnd-kit/modifiers": "^1.0.1",
     "@dnd-kit/sortable": "^1.0.1",
+    "@radix-ui/react-aspect-ratio": "^0.0.14",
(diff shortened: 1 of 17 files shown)


milg15:
The preview and the image that it's being saved doesn't save/show in preview the borders 🤔  
image
image

milg15 (commented):
Everything in the code looks good 👍  The borders error it's the only thing I saw that works weird. I am not 100% if that's how it's supposed to work.

dabbott:
Good catch! I didn't realize that we need to adjust the frame for borders and also trim whitespace off images. I'm going to merge this now and come back to that

PR #130, opened 2021-06-26
Render image fills on the canvas

milg15:
Render Pattern preview


packages/noya-renderer/src/primitives.ts
```diff
@@ -163,8 +165,120 @@ export function fill(
...
+                CanvasKit.Matrix.scaled(scale, scale),
+              ),
+            ),
+          );
+          break;
+        }
+        case Sketch.PatternFillType.Stretch: {
+          paint.setShader(
+            canvasImage.makeShaderCubic(
+              CanvasKit.TileMode.Clamp,
```
dabbott, line 230: I noticed the other 2 are `Decal` and this one is `Clamp` - is it different in this case? If not, let's use `Decal` just for consistency

packages/noya-renderer/src/primitives.ts
```diff
@@ -163,8 +165,120 @@ export function fill(
...
+          const scale = Math.max(
+            scaledRect.width / canvasImage.width(),
+            scaledRect.height / canvasImage.height(),
+          );
+
+          paint.setShader(
+            canvasImage.makeShaderCubic(
+              CanvasKit.TileMode.Decal,
+              CanvasKit.TileMode.Decal,
+              1,
```
dabbott, line 214: Does the 1 help here?

dabbott (approved):
Looks great! 

I think after this we can refactor some of the scaling stuff into the `resize` util so it's easier to re-use. I didn't realize that we'd pretty much always need to know the scale when we use `resize`. After that, I think we can combine all 3 cases, fill/fit/stretch, into one.

packages/noya-renderer/src/primitives.ts
milg15: I am not sure, I'll leave it to 0 for consistency.

PR #132, opened 2021-06-27
Draggable edit points

### LINKED ISSUE #116: Draggable points and control points
Clicking and dragging a point or control point should change its coordinates.

- Since multiple points may be selected, we should be able to drag multiple points simultaneously. Clicking and dragging _any_ selected point should move _all_ selected points by the same delta (just like dragging multiple layers)
- Only one control point may be selected, so only one may be moved at a time
- Dragging a `Mirrored` or `Asymmetric` control point should also update the opposite control point.

> Reminder: Mirrored control points have the same slope and distance, while asymmetric control points may have a different distance.

## Implementation

This should be analogous to dragging layers. We should add new actions and states to the `interactionReducer`, and handle mouse clicks in `Canvas.tsx`. Our new states will probably be:

- For points: `maybeMovePoint`, `movingPoint`
- For control points: `maybeMoveControlPoint`, `movingControlPoint`

We should have corresponding actions for these states. **We should try to follow the same naming/patterns as for moving layers, where possible**.

When a point is clicked, we should enter a `maybe` state, where that click might turn into a drag:

https://github.com/noya-app/noya/blob/ba2acae97211474540a1f061b18a447eb0adf24c/packages/app/src/containers/Canvas.tsx#L328

From there, if we move greater than 2px in either the x or y axis, we should enter a moving state:

https://github.com/noya-app/noya/blob/ba2acae97211474540a1f061b18a447eb0adf24c/packages/app/src/containers/Canvas.tsx#L359-L378

We should handle updating state in the `canvasReducer`, in a case similar to our current `moving`. (Snapping is out of scope for this issue)

https://github.com/noya-app/noya/blob/ba2acae97211474540a1f061b18a447eb0adf24c/packages/noya-state/src/reducers/canvasReducer.ts#L161

When the mouse is lifted, reset the interaction state:

https://github.com/noya-app/noya/blob/ba2acae97211474540a1f061b18a447eb0adf24c/packages/app/src/containers/Canvas.tsx#L546-L548

leannepepper:
Adding ability to drag both select and control points in pathEdit mode

closes #116


packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -223,6 +257,230 @@ export function canvasReducer(
...
+            const boundingRects = getBoundingRectMap(
+              getCurrentPage(state),
+              Object.keys(state.selectedPointLists),
+              {
+                clickThroughGroups: true,
+                includeArtboardLayers: false,
+                includeHiddenLayers: false,
+              },
+            );
+            //Find point to measure distance
```
leannepepper, line 272: I feel like there could be a better way to calculate the distance between drag start and drag end so we don't have to pass a selectedPoint through the action, any ideas?

dabbott: We can try to use `origin` and `pageSnapshot`, like we do for moving layers.

This approach is potentially safer anyway, in case we encounter a degenerate case where the layer's frame has width/height of 0 after computing the tight bounds, since we scale the points by the layer's frame. I'm not sure if that can occur - it can't occur when moving layers, but it can for scaling them, which we do here

dabbott (changes requested):
Looks great! I found a couple things we can try to refactor to reduce duplicated code

packages/app/src/containers/Canvas.tsx
```diff
@@ -71,6 +74,12 @@ function getPoint(event: MouseEvent): Point {
   return { x: Math.round(event.offsetX), y: Math.round(event.offsetY) };
 }
 
+function isMoving(point: Point, origin: Point): boolean {
+  if (Math.abs(point.x - origin.x) > 2 || Math.abs(point.y - origin.y) > 2) {
```
dabbott, line 78: In cases like this, with a single if/else that returns a bool, we can usually return the expression directly. E.g. `return Math.abs(...) || ...`

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -148,7 +168,8 @@ export function canvasReducer(
 
         switch (interactionState.type) {
           case 'editPath': {
-            //Selects the first point in the first selected layer and initializes a point list for each selected layer
+            if (action[1][0] === 'resetEditPath') break;
```
dabbott, line 171: We should put back that comment on the `forEach` that we took out

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -97,203 +91,46 @@ export function pointReducer(
...
-          };
-
-          // Transform back to the range [0, 1], using the new bounds
-          const encodedPoints = decodedPoints.map((decodedCurvePoint) =>
-            encodeCurvePoint(decodedCurvePoint, layer.frame),
-          );
-
-          layer.points = encodedPoints;
-        });
+        if (!draft.selectedControlPoint) return state;
```
dabbott, line 121: We don't need to return `state` within an immer `produce`. It can just be `return`.

We can also remove `if (!state.selectedControlPoint) return state;` above if we check down here

packages/noya-state/src/selectors/pointSelectors.ts
```diff
@@ -84,3 +97,209 @@ export const getSelectedControlPoint = (
...
+      ...layer.frame,
+      ...newRect,
+    };
+
+    // Transform back to the range [0, 1], using the new bounds
+    draftLayer.points = decodedPoints.map((decodedCurvePoint) =>
+      encodeCurvePoint(decodedCurvePoint, draftLayer.frame),
+    );
+  });
+};
```
dabbott, line 191: Do we wanna try to make this a separate function? It's ok if we leave it for another time too

dabbott (approved):
I found a couple tiny things but LGTM

PR #134, opened 2021-06-28
Refactor fill picker

dabbott:
This is mainly a refactor of the `FillInputFieldWithPicker`, and everything that renders it. I improved the "prop drilling" and optional props issues by bundling props together into groups:

- `colorProps`
- `gradientProps`
- `patternProps`

This also fixes an issue where we showed invalid fill types for shadows and borders. E.g. shadows can only accept color fills.

Aside from that, some misc UI cleanup.

This refactor should make it easier to support inspecting multiple borders and fills, which I plan to add next.


milg15 (approved):
It looks greatt!! No more optional functions 🎉

PR #135, opened 2021-06-29
add trackpad panning

souporserious:
Sets up [react-use-gesture](https://use-gesture.netlify.app/) to handle panning with a trackpad.

### DIFF
--- a/packages/app/src/containers/Canvas.tsx
+++ b/packages/app/src/containers/Canvas.tsx
@@ -1,5 +1,5 @@
 import Sketch from '@sketch-hq/sketch-file-format-ts';
-import { ContextMenu } from 'noya-designsystem';
+import { ContextMenu, mergeEventHandlers } from 'noya-designsystem';
 import { createRect, Insets } from 'noya-geometry';
 import { useKeyboardShortcuts } from 'noya-keymap';
 import { useCanvasKit, uuid } from 'noya-renderer';
@@ -32,9 +32,8 @@ import {
   useMemo,
   useRef,
 } from 'react';
-import styled, { useTheme } from 'styled-components';
 import { useGesture } from 'react-use-gesture';
-import mergeProps from 'merge-props';
+import styled, { useTheme } from 'styled-components';
 import {
   useApplicationState,
   useSelector,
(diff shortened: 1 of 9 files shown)


packages/app/src/containers/Canvas.tsx
```diff
@@ -106,6 +108,17 @@ export default memo(function Canvas() {
   const containerSize = useSize(containerRef);
   const meta = useSelector(Selectors.getCurrentPageMetadata);
   const { setCanvasSize, highlightLayer, highlightedLayer } = useWorkspace();
+  const bind = useGesture({
+    onWheel: ({ movement: [x, y] }) => {
+      const rawPoint = { x, y };
+      dispatch('interaction', [
+        state.interactionState.type === 'maybePan'
```
souporserious, line 115: @dabbott this isn't working right now. I had something similar to what you have below, but that was interfering with the current panning approach. Should we add a specific case for trackpad panning in the reducer or should we be able to tweak `updatePanning` to work here?

dabbott: I'm leaning toward a separate action. The current panning is stateful: you have to be in panMode, and then go from maybePan to panning, and there are several actions to coordinate that. Panning via gesture should theoretically be stateless... i.e. it could all happen with a single `pan` action that moves the canvas by a delta. It wouldn't need to go through `interactionState`, since you can always pan via trackpad even when you're in another state.

packages/app/src/containers/Canvas.tsx
```diff
@@ -98,6 +99,11 @@ export default memo(function Canvas() {
   const containerSize = useSize(containerRef);
   const meta = useSelector(Selectors.getCurrentPageMetadata);
   const { setCanvasSize, highlightLayer, highlightedLayer } = useWorkspace();
+  const bind = useGesture({
+    onWheel: ({ delta: [x, y] }) => {
+      dispatch('gesturePan', { x, y });
```
souporserious, line 104: Should we be more specific and call the action `wheelPan`?

dabbott: Or maybe more general and call it `pan`. We already have `panning`, `maybePan`, `panMode`, etc, without any qualifiers, so I think that'd be more consistent. Maybe at some point I'll update those to use this new `pan`

dabbott (approved):
This is amazing! It's so little code and it "just works"

PR #137, opened 2021-06-29
Theme Selectors

milg15:


packages/noya-state/src/reducers/themeReducer.ts
```diff
@@ -307,32 +311,19 @@ export function themeReducer(
...
-        updateSelection(draft.selectedTextStyleIds, id, selectionType);
-      });
-    }
+    case 'selectSwatch':
+    case 'selectThemeStyle':
+    case 'selectTextStyle':
     case 'selectSymbol': {
       const [, id, selectionType = 'replace'] = action;
 
+      const crrThemeTab = state.currentThemeTab;
```
dabbott, line 320: Rename `crrThemeTab` => `currentThemeTab`

We don't abbreviate names much, and good to keep the names consistent with what we already have

packages/noya-state/src/reducers/applicationReducer.ts
```diff
@@ -57,14 +58,7 @@ export type ApplicationState = {
...
   selectedControlPoint?: SelectedControlPoint;
-  selectedSwatchIds: string[];
-  selectedLayerStyleIds: string[];
-  selectedTextStyleIds: string[];
-  selectedSymbolsIds: string[];
-  selectedSwatchGroup: string;
-  selectedTextStyleGroup: string;
-  selectedThemeStyleGroup: string;
-  selectedSymbolGroup: string;
+  selectedTheme: Record<ThemeTab, ThemeSelection>;
```
dabbott, line 61: Let's rename `selectedTheme` => `selectedThemeTab`. `selectedTheme` makes me think light/dark theme

dabbott (approved):
Looks great! I think this is a nice improvement 😁

PR #138, opened 2021-06-29
Improves smaller  Pattern Previews

milg15:
Uses URL.createObjectURL over hardcoded encoding option. It also allows show in the canvas the change in opacity for gradients and patterns.


packages/app/src/components/inspector/PatternInspector.tsx
```diff
@@ -38,6 +39,7 @@ const Container = styled.div<{
   backgroundSize,
   imageRendering: 'crisp-edges',
   width: '100%',
+  transform: 'scale(0.999)',
```
dabbott, line 42: What does this do? I'm guessing it forces 3d rendering, but what does it do here?

packages/app/src/hooks/useObjectURL.ts
```diff
@@ -0,0 +1,17 @@
+import { useLayoutEffect, useMemo } from 'react';
+
+export function useObjectURL(object: string | ArrayBuffer | Uint8Array | Blob) {
+  const objectURL = useMemo(() => {
+    if (typeof object === 'string') return object;
```
dabbott, line 5: We probably don't really want strings here, cause we don't use `createObjectURL`. Or at least we shouldn't call `revokeObjectURL` if the object is a string. If we make the change below about default patterns, we may not need to handle strings here (yet - Sketch does support base64 data, but we can come back to that).

packages/noya-designsystem/src/utils/sketchPattern.ts
```diff
@@ -20,12 +19,9 @@ export function getPatternBackground(
   image?: Sketch.FileRef | Sketch.DataRef,
 ) {
   if (!image || !images[image._ref])
-    return `url(
-        'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYAQMAAADaua+7AAAABlBMVEUAAAAAAAClZ7nPAAAAAXRSTlMAQObYZgAAABNJREFUCNdjYOD/TxL+/4GBFAwAvMsj3bQ3H74AAAAASUVORK5CYI=A',
-      )`;
+    return `'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYAQMAAADaua+7AAAABlBMVEUAAAAAAAClZ7nPAAAAAXRSTlMAQObYZgAAABNJREFUCNdjYOD/TxL+/4GBFAwAvMsj3bQ3H74AAAAASUVORK5CYI=A',`;
```
dabbott, line 22: Let's import `CHECKERED_BACKGROUND_BYTES` here, so we don't have to handle strings later.

We just need to export it from here: https://github.com/noya-app/noya/blob/ad1cee93813e33f299b2286ec39d4e979104e83b/packages/noya-renderer/src/hooks/useCheckeredFill.ts#L15

packages/noya-designsystem/src/components/FillInputField.tsx
```diff
@@ -39,6 +40,13 @@ export default forwardRef(function FillInputField(
   const [state] = useApplicationState();
   const { inputBackground, placeholderDots } = useTheme().colors;
 
+  const backgroundPattern = useObjectURL(
```
dabbott, line 43: We might need to move this call into a wrapper component, since we don't really want to call it for things that aren't patterns

packages/app/src/components/inspector/PickerPatterns.tsx
```diff
@@ -2,15 +2,29 @@ import type Sketch from '@sketch-hq/sketch-file-format-ts';
...
-  onChange: (color: Sketch.FileRef | Sketch.DataRef) => void;
+  onChange: (file: Sketch.FileRef | Sketch.DataRef) => void;
 }
 
+const PatternSquare = memo(
+  ({
+    image,
+    onClick,
+  }: {
+    image: ArrayBuffer | string;
```
dabbott, line 20: I think we do want to pass `item` in here after all, so that `onClick` doesn't need to be created inline below. E.g.

```tsx
<Square onClick={useCallback(() => onClick(item)} />
```

This will currently not get memoized, since we redefine `onClick` every render of `<PatternSquare onClick={() => onChange(item)} />`

packages/noya-designsystem/src/utils/sketchPattern.ts
dabbott: I realized there's an issue with this default image approach: we don't put the image data into the `FileMap`, so it won't get rendered by the canvas. We should probably set all that up correctly when we add the `image` object to the fill: 

https://github.com/noya-app/noya/blob/ad1cee93813e33f299b2286ec39d4e979104e83b/packages/noya-state/src/reducers/styleReducer.ts#L327

We may need to handle some portion of this in the `applicationReducer` so that we have access to `sketch.images`

dabbott: It's actually valid to have a pattern with no image - Sketch allows deleting the image. So we may need to handle `undefined` as a return value from this function, or not call it in the first place if there's no `image`

dabbott (changes requested):
Looks good! I also found an issue with pattern rendering that I didn't catch before, but I think now's a good time to solve it

packages/app/src/components/inspector/PatternInspector.tsx
milg15: that was a test

packages/noya-designsystem/src/utils/sketchPattern.ts
milg15: wait so I write the placeholder checkered_background in the image on creation? and what should it show when you don't have any _ref ? Just all white?

milg15: Okey so I should add this CHECKERED_BACKGROUND to the images if it doesn't exist, i think i got it

packages/noya-designsystem/src/components/FillInputField.tsx
dabbott: I just pushed a fix for this - I made the background fill a separate component to solve this issue

packages/noya-state/src/reducers/applicationReducer.ts
```diff
@@ -195,8 +196,17 @@ export function applicationReducer(
...
               layer.style = styleReducer(layer.style, action);
+
+              if (
+                action[0] === 'setFillFillType' &&
+                action[2] === Sketch.FillType.Pattern &&
+                layer.style.fills
+              ) {
+                const _ref = layer.style.fills[action[1]].image?._ref;
+                if (!_ref) return;
+                draft.sketch.images[_ref] = CHECKERED_BACKGROUND_BYTES;
```
dabbott, line 208: Currently, this will replace the image if the user chooses another fill type, e.g. gradient, and then sets it back to pattern

We might want to do this logic before we call `styleReducer()`. We could even do it separately from the big list of cases (`addNewBorder`, `addNewFill`,...) as a top-level case, rather than calling into the `styleReducer`, if that makes things easier

dabbott (changes requested):
I added some commits to this branch to refactor the pattern preview. I wanted the `useObjectURL` to only be called for patterns specifically, but I wasn't quite sure the best way to do it. I made some new components and I think it works pretty well, but I might not have caught every case.

I deleted our special .pdf handling, since I think that it's simpler to just not support pdfs for now.

packages/noya-renderer/src/primitives.ts
```diff
@@ -170,72 +171,87 @@ export function fill(
...
-                  fill.patternTileScale,
-                  fill.patternTileScale,
-                ),
-              ),
-            ),
-          );
+        let canvasImage = CanvasKit.MakeImageFromEncoded(image);
+        canvasImage = canvasImage
+          ? canvasImage
+          : CanvasKit.MakeImageFromEncoded(CHECKERED_BACKGROUND_BYTES);
```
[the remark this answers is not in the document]
milg15: We do to be able to show .sketch files that already have files that don't have this _ref but are new Fill with patterns. Like in the case of the current file I'm using for testing.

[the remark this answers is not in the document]
dabbott: If it's just a matter of not being able to decode pdfs, I don't think we need to support that case, so let's take this out. The important thing is that we can create new fill patterns in Noya.

We can probably add pdf support later on via https://github.com/mozilla/pdf.js to cover that case

packages/noya-state/src/reducers/applicationReducer.ts
```diff
@@ -218,6 +237,16 @@ export function applicationReducer(
...
             style.value = styleReducer(style.value, action);
 
+            if (
+              action[0] === 'setFillFillType' &&
+              action[2] === Sketch.FillType.Pattern &&
+              style.value.fills
+            ) {
+              const _ref = style.value.fills[action[1]].image?._ref;
+              if (!_ref) return;
+              draft.sketch.images[_ref] = CHECKERED_BACKGROUND_BYTES;
```
dabbott, line 247: Should this pattern initialization the new code above? Maybe we should make a function so we don't have it duplicated twice

dabbott (changes requested):

packages/noya-state/src/reducers/applicationReducer.ts
milg15: which code ? 🤔

milg15: oh I understand !

packages/noya-state/src/selectors/themeSelectors.ts
```diff
@@ -142,3 +144,19 @@ export const getSymbolsInstancesIds = (
     ),
   );
 };
+
+export const setNewPaternFill = (
```
dabbott, line 148: `setNewPaternFill` => `setNewPatternFill`

dabbott (approved):

PR #143, opened 2021-07-03
Handle other Export Formats

milg15:


packages/app/src/containers/ExportInspector.tsx
```diff
@@ -45,26 +50,117 @@ export default memo(function ExportInspector() {
...
+  const handleFileFormat = useCallback(
+    async (exportFormat: Sketch.ExportFormat) => {
+      const {
+        scale,
+        absoluteSize,
+        fileFormat,
+        visibleScaleType,
+      } = exportFormat;
+
+      const adjustSize = {
```
dabbott, line 62: This size logic looks good, but let's move it to a separate function and use a `switch`, so it's clearer how width and height are related. E.g.

```ts
function getExportSize(exportFormat: Sketch.ExportFormat, size: Size) {
  switch (visibleScaleType) {
    case Sketch.VisibleScaleType.Width:
      return { width: ..., height: ...}
    case ...
  }
}
```

```ts
const size = getExportSize(exportFormat, selectedLayer.frame)
```

packages/app/src/containers/ExportInspector.tsx
```diff
@@ -17,15 +17,20 @@ import {
...
 
-async function saveFile(name: string, data: Uint8Array) {
+async function saveFile(
+  name: string,
+  fileFormat: string,
+  data: Uint8Array | ArrayBuffer,
+) {
   const file = new File([data], name, {
-    type: 'image/png',
+    type: `${fileFormat === 'zip' ? 'application' : 'image'}/${fileFormat}`,
```
dabbott, line 28: We'll leave this for now, but I'm going to add some file type stuff to utils and will change this then

packages/app/src/containers/ExportInspector.tsx
```diff
@@ -45,26 +50,117 @@ export default memo(function ExportInspector() {
...
+    (exportFormat: Sketch.ExportFormat) => {
+      const {
+        scale,
+        name,
+        namingScheme,
+        fileFormat,
+        visibleScaleType,
+      } = exportFormat;
+
+      return `${!namingScheme && name ? `${name}.` : ''}${selectedLayer.name}${
```
dabbott, line 103: We don't need to include the `scale` or `visibleScaleType` in the name. So the logic will be:

- if there's no naming scheme: return `${selectedLayer.name}.${fileFormat}`
- if the naming scheme is Prefix: `${name}${selectedLayer.name}.${fileFormat}`
- if the naming scheme is Suffix: `${selectedLayer.name}${name}.${fileFormat}`

packages/app/src/containers/ExportInspector.tsx
```diff
@@ -17,15 +17,20 @@ import {
 } from '../contexts/ApplicationStateContext';
 import useCanvasKit from '../hooks/useCanvasKit';
 import { renderImageFromCanvas } from '../utils/renderImageFromCanvas';
+import JSZip from 'jszip';
 
-async function saveFile(name: string, data: Uint8Array) {
+async function saveFile(
+  name: string,
+  fileFormat: string,
+  data: Uint8Array | ArrayBuffer,
```
dabbott, line 25: The type here can be just `ArrayBuffer`. If we need to get an `ArrayBuffer` from a `Uint8Array`, we can use `uint8Array.buffer`.

packages/app/src/containers/ExportInspector.tsx
```diff
@@ -45,26 +50,117 @@ export default memo(function ExportInspector() {
   const exportFormats = selectedLayer.exportOptions.exportFormats;
 
   // TODO: Handle export formats
+  const handleFileFormat = useCallback(
+    async (exportFormat: Sketch.ExportFormat) => {
+      const {
+        scale,
+        absoluteSize,
```
dabbott, line 57: I'm now wondering... do we need to update `absoluteSize` whenever we scale a layer? 🤔  Otherwise I think `absoluteSize` will be stale. We could come back to this.

dabbott (changes requested):
Looks great! 💯

dabbott (approved):

PR #145, opened 2021-07-06
Insert SymbolInstance into the current page

milg15:


dabbott:
This is a good start!

Instead of using a cursor image (I'm assuming that's what `getBackgroundImageCursor` would do), we should render the symbol preview in `<SketchFileRenderer>` when we're in the `insertSymbol` state. Let me know if you have questions about how to do that.

I'd recommend rebasing on main soon since there are a lot of new changes to sync:

```
git checkout main
git pull
git checkout symbols-into-page
git rebase main
```

milg15:
Oh okey this rebase it's getting too crazy 😆 I'll create a new branch instead. This one it's getting a bit too messy and I haven't made that many changes yet.

PR #147, opened 2021-07-07
adding support for closing path when clicking on qualified point

leannepepper:
The cursor 'no-drop' will appear when hovering over a point that, when clicked, will close the path of an open layer.

### DIFF
--- a/packages/app/src/containers/Canvas.tsx
+++ b/packages/app/src/containers/Canvas.tsx
@@ -123,6 +123,11 @@ export default memo(function Canvas() {
     'Shift-ArrowDown': () => nudge('Y', 10),
     Backspace: () => dispatch('deleteLayer', state.selectedObjects),
     Escape: () => dispatch('interaction', ['reset']),
+    Shift: () => dispatch('setKeyModifier', 'shiftKey', true),
+  });
+
+  useKeyboardShortcuts('keyup', {
+    Shift: () => dispatch('setKeyModifier', 'shiftKey', false),
   });
 
   const insets = useMemo(
@@ -230,11 +235,8 @@ export default memo(function Canvas() {
           dispatch('interaction', ['editPath']);
           break;
         }
-        case 'editPath':
-        case 'movingControlPoint':
-        case 'movingPoint':
-        case 'maybeMoveControlPoint':
-        case 'maybeMovePoint': {
+        case 'editPath': {
+          const { shiftKey } = state.keyModifiers;
           let selectedPoint: SelectedPoint | undefined = undefined;
           let selectedControlPoint: SelectedControlPoint | undefined;
 
@@ -247,6 +249,7 @@ export default memo(function Canvas() {
               includeHiddenLayers: false,
             },
           );
+
           getSelectedLayers(state)
             .filter(Layers.isPointsLayer)
             .forEach((layer) => {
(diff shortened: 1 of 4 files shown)


packages/noya-state/src/selectors/elementSelectors.ts
```diff
@@ -21,12 +21,30 @@ type PathElement =
       value: SelectedControlPoint;
     };
 
+export function canClosePath(state: ApplicationState, element: PathElement) {
+  if (element.type === 'point') {
+    const [layerId, index] = element.value;
+
+    const layers = getSelectedLayers(state)
```
dabbott, line 28: This should work, but rather than looking up an array of layers, filtering it down, and picking the 0 element, we can find a layer directly and more safely via:

```ts
const layer = Layers.find(getCurrentPage(state), (layer) => layer.do_objectID === layerId);

if (!layer || !Layers.isPointsLayer(layer)) return false
```

If we do this in maybe 1 more place, we can make a `findLayerInCurrentPage` util

packages/app/src/containers/Canvas.tsx
```diff
@@ -287,6 +290,11 @@ export default memo(function Canvas() {
                 : 'replace',
             );
             dispatch('interaction', ['maybeMovePoint', point]);
+            const canClose = canClosePath(state, {
+              type: 'point',
+              value: selectedPoint,
+            });
+            if (canClose) dispatch('setIsClosed', true);
```
dabbott, line 297: We should also check that shift isn't held, since that should add a point to the selection without closing the path

packages/noya-state/src/selectors/elementSelectors.ts
```diff
@@ -21,12 +21,30 @@ type PathElement =
...
+      .filter(Layers.isPointsLayer)
+      .filter((layer) => layer.do_objectID === layerId);
+
+    return (
```
dabbott, line 32: This is close! We only want to close the path if the first/last point is selected, but at the moment this lets us close the path if _any_ other point is selected.

It does a little extra work, but it'd probably be best to use `getIndexPathOfOpenShapeLayer` to determine if we're in a scenario where we can close a path, since we know that works correctly and then we'll have consistent behavior everywhere.

```ts
// A path needs at least 2 points to be closed.
// TODO: Don't allow closing a straight line segment with 2 points.
if (layer.points.length < 2) return false

const selectedPoint = getIndexPathOfOpenShapeLayer(state)

if (!selectedPoint) return false

const lastIndex = layer.points.length - 1

return (
  (index === lastIndex && selectedPoint.pointIndex === 0) ||
  (index === 0 && selectedPoint.pointIndex === lastIndex)
)
```

I threw in a TODO for the special case of a 2-point path, but we don't need to handle that right now

dabbott (changes requested):
Good thinking with the small PR! There are definitely a lot of cases to consider here, which makes it pretty tricky.

packages/app/src/containers/Canvas.tsx
leannepepper: Currently, the pointer will still change to 'no-drop' even if shift is true (but it will not close the path). Should we have another interaction state to handle this use case?

dabbott: Good idea! I added a `keyModifiers` to `ApplicationState` to solve this issue.

dabbott (approved):

PR #148, opened 2021-07-07
Add symbol to page

milg15:
Allows to select a symbol an drop it into the canvas

### DIFF
--- a/packages/noya-renderer/src/components/SketchFileRenderer.tsx
+++ b/packages/noya-renderer/src/components/SketchFileRenderer.tsx
@@ -1,3 +1,5 @@
+import type Sketch from '@sketch-hq/sketch-file-format-ts';
+
 import { useApplicationState } from 'app/src/contexts/ApplicationStateContext';
 import { useWorkspace } from 'app/src/hooks/useWorkspace';
 import * as CanvasKit from 'canvaskit';
@@ -24,6 +26,7 @@ import {
   getScreenTransform,
   getSelectedLayerIndexPaths,
   getSelectedLayerIndexPathsExcludingDescendants,
+  getSymbols,
 } from 'noya-state/src/selectors/selectors';
 import {
   getAxisValues,
@@ -451,6 +454,26 @@ export default memo(function SketchFileRenderer() {
     );
   }, [isEditingPath, page, state]);
 
+  const symbol = useMemo(() => {
+    if (interactionState.type !== 'insertingSymbol') return;
+    // TODO: Modifying only the instance of the symbol master, not the master itself. Improve mouse pointer position.
+    const symbol = {
+      ...getSymbols(state).find(
+        ({ do_objectID }) => do_objectID === interactionState.symbolID,
+      ),
+    } as Sketch.SymbolMaster;
+
+    if (!symbol || !symbol.style) return;
+    symbol.style.contextSettings = {
+      _class: 'graphicsContextSettings',
+      blendMode: 1,
+      opacity: 0.4,
+    };
+    symbol.frame = { ...symbol.frame, ...interactionState.point };
+
+    return symbol;
+  }, [state, interactionState]);
+
   return (
     <>
       <RCKRect rect={canvasRect} paint={backgroundFill} />
@@ -493,6 +516,9 @@ export default memo(function SketchFileRenderer() {
         )}
       </Group>
       <Group transform={screenTransform}>
+        {interactionState.type === 'insertingSymbol' && symbol && (
+          <SketchLayer key={symbol.symbolID} layer={symbol} />
+        )}
         {interactionState.type === 'marquee' && (
           <Marquee
             rect={createRect(interactionState.origin, interactionState.current)}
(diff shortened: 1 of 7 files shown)


packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -451,6 +454,26 @@ export default memo(function SketchFileRenderer() {
...
+    if (interactionState.type !== 'insertingSymbol') return;
+    // TODO: Modifying only the instance of the symbol master, not the master itself. Improve mouse pointer position.
+    const symbol = {
+      ...getSymbols(state).find(
+        ({ do_objectID }) => do_objectID === interactionState.symbolID,
+      ),
+    } as Sketch.SymbolMaster;
+
+    if (!symbol || !symbol.style) return;
+    symbol.style.contextSettings = {
```
dabbott, line 467: We can avoid modifying the original symbol master via `produce`:

```ts
produce(symbol, (draft) => {
  symbol.style.contextSettings = ...
})
```

dabbott (commented):
Looks good so far!

packages/noya-renderer/src/components/SketchFileRenderer.tsx
milg15: ohh! I'll do that

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -493,6 +516,9 @@ export default memo(function SketchFileRenderer() {
         )}
       </Group>
       <Group transform={screenTransform}>
+        {interactionState.type === 'insertingSymbol' && symbol && (
+          <SketchLayer key={symbol.symbolID} layer={symbol} />
```
dabbott, line 520: We may want to render `<SketchArtboardContent />` directly instead of `<SketchLayer />`, so that we don't show the symbol's title

Screen Shot 2021-07-08 at 12 54 37 PM

Alternately we could instantiate a symbolInstance, but that seems like more work

milg15: It's easier when using the SketchArtboardContent, it did something weird when using symbolInstance

milg15:
I just noticed I didn't push a change I did earlier today 😆 The interaction-canvas reducer interaction it's weird. I am not 100% it's the best way of calling that them

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -535,6 +538,43 @@ export function canvasReducer(
 
             break;
           }
+          case 'addSymbolLayer': {
```
dabbott, line 541: We should move `'addSymbolLayer'` into the `canvasReducer`, since it doesn't depend on the interaction state. I.e. we can define the action next to `'addShapePathLayer'`, and we can move this `case` into the top-level `switch` of the canvasReducer. 

That should simplify things a little, and it'll also let us move the `uuid()` call directly into the canvasReducer, so we can remove the `id` parameter from the action

dabbott (changes requested):
This is really great! It makes Noya feel a lot more like a real design tool 😄

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -535,6 +538,43 @@ export function canvasReducer(
...
+
+            if (parent && Layers.isChildLayer(layer)) {
+              layer.frame.x -= parent.frame.x;
+              layer.frame.y -= parent.frame.y;
+
+              parent.layers.push(layer);
+            } else {
+              draft.sketch.pages[pageIndex].layers.push(layer);
+            }
+          }
```
dabbott, line 577: Let's also automatically select the newly created layer, `draft.selectedObjects = [layer.do_objectID]`

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -451,11 +455,43 @@ export default memo(function SketchFileRenderer() {
...
+
+      draft.style.contextSettings = {
+        _class: 'graphicsContextSettings',
+        blendMode: 1,
+        opacity: 0.5,
+      };
+
+      draft.frame = {
+        ...symbol.frame,
+        x: interactionState.point!.x - symbol.frame.width / 2,
```
dabbott, line 479: I think we can avoid the `point!` if we assign it to a variable outside the `produce` callback. E.g. `const point = interactionState.point`. TypeScript "forgets" that we already checked for `!interactionState.point` once we're in a different function

packages/noya-state/src/selectors/layerSelectors.ts
```diff
@@ -120,3 +120,24 @@ export function findSymbolMaster<T extends Sketch.SymbolMaster | undefined>(
     (child) => Layers.isSymbolMaster(child) && symbolID === child.symbolID,
   ) as T;
 }
+
+export function addToParentLayer(
```
dabbott, line 124: 👍 Good idea!

dabbott (approved):

PR #153, opened 2021-07-09
Pen tool control points

leannepepper:
Adding ability to drag control points when creating a path with pen tool.


packages/noya-renderer/src/components/PseudoPathLine.tsx
```diff
@@ -1,25 +1,50 @@
...
   const points = useMemo(() => {
-    return [point, decodedCurvePoint.point];
-  }, [decodedCurvePoint.point, point]);
+    return [encodedPointToDraw, curvePoint];
+  }, [encodedPointToDraw, curvePoint]);
+
+  const path = Primitives.path(CanvasKit, points, frame, false);
+
+  const strokedPath = useMemo(
+    () => Primitives.getStrokedBorderPath(CanvasKit, path, 1, 0),
```
leannepepper, line 45: I've hardcoded the values for the borderPath with and position, is there a use case where these values would need to be different?

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -407,16 +407,13 @@ export default memo(function SketchFileRenderer() {
...
-      layer.frame,
-    );
 
     return (
       <>
         <PseudoPathLine
           point={interactionState.point}
-          decodedCurvePoint={decodedCurvePoint}
+          curvePoint={layer.points[indexPath.pointIndex]}
+          frame={layer.frame}
```
leannepepper, line 416: Would it be better to pass in the entire layer and indexPath rather than the curvePoint and frame?

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -188,17 +187,18 @@ export function canvasReducer(
...
+        const decodedPoint: DecodedCurvePoint = {
           _class: 'curvePoint',
           cornerRadius: 0,
-          curveFrom: stringifyPoint({ x: 0, y: 0 }),
-          curveTo: stringifyPoint({ x: 0, y: 0 }),
-          hasCurveFrom: false,
-          hasCurveTo: false,
+          curveFrom: point,
+          curveTo: point,
+          hasCurveFrom: true,
```
leannepepper, line 195: I have these curveFrom and curveTo set to true by default to render the control point curves. Would this be better as a point action in the point reducer?

dabbott: Yeah, I think so, otherwise we won't be able to do the right thing when switching the point's curve mode for the first time. I think it can fit into our existing actions, so we won't need to make a new one.

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -241,10 +241,10 @@ export function canvasReducer(
         const decodedPoint: DecodedCurvePoint = {
           _class: 'curvePoint',
           cornerRadius: 0,
-          curveFrom: { x: 0, y: 0 },
-          curveTo: { x: 0, y: 0 },
-          hasCurveFrom: false,
-          hasCurveTo: false,
+          curveFrom: point,
+          curveTo: point,
+          hasCurveFrom: true,
```
dabbott, line 246: These should also start false, since we have a straight point to begin with.

packages/app/src/containers/Canvas.tsx
```diff
@@ -433,6 +435,25 @@ export default memo(function Canvas() {
...
+
+          if (isMoving(point, origin)) {
+            dispatch('setPointCurveMode', Sketch.CurveMode.Mirrored);
+            dispatch(
+              'selectControlPoint',
+              selectedLayers[0].do_objectID,
+              0,
+              'curveFrom',
+            );
+            dispatch('interaction', ['maybeMoveControlPoint', origin]);
```
dabbott, line 449: This `maybe` state isn't really necessary here, so I'll push a quick fix to remove it after we merge this PR. I might have to change things in a couple other places though, so I'll do it separately.

packages/noya-renderer/src/components/PseudoPathLine.tsx
```diff
@@ -1,25 +1,50 @@
...
+    point,
+  };
+
+  const encodedPointToDraw = encodeCurvePoint(decodedPointToDraw, frame);
+
   const points = useMemo(() => {
-    return [point, decodedCurvePoint.point];
-  }, [decodedCurvePoint.point, point]);
+    return [encodedPointToDraw, curvePoint];
+  }, [encodedPointToDraw, curvePoint]);
```
dabbott, line 40: This memoization won't work as we have it, since we create a new point each render. We can move the `decodedPointToDraw` and `encodedPointToDraw` into the `useMemo` that should work. The `path` can probably go into the same `useMemo`

dabbott (changes requested):
🔥 Looks great!

I still get a crash every once in a while, or the state gets weird when adding new paths, but we can come back to those. I don't think they're introduced in this PR.

packages/noya-renderer/src/components/PseudoPathLine.tsx
leannepepper: refactored to have this component to take a CurvePoint[]

dabbott (approved):

PR #156, opened 2021-07-10
Drag & drop image

milg15:
Allows to drag & drop an image into the canvas 

| Drag | Drop |
| ------------- | ------------- |
|  |


packages/app/src/containers/Canvas.tsx
```diff
@@ -711,11 +713,58 @@ export default memo(function Canvas() {
...
+
+      if (!extension) {
+        alert(
+          `Files of type ${file.type} aren't supported. Files of type png, jpg, and webp are supported.`,
+        );
+        return;
+      }
+
+      const data = await file.arrayBuffer();
+      const img = new Image();
```
dabbott, line 739: This approach is totally fine but... I think I wanna use CanvasKit for consistency after all. Plus, in the future we can share the decoded version in `ImageCache` so we only have to decode once.

```ts
const image = CanvasKit.MakeImageFromEncoded(data);

if (!image) return;

const width = image.width();
const height = image.height();

image.delete();
```

Also note that if we did go with the `createObjectURL` approach, we'd want to revoke the URL after measuring

dabbott (commented):
Looks good so far!

packages/app/src/containers/Canvas.tsx
milg15: I forgot about that function 😅 !

PR #167, opened 2021-07-12
Drop image to canvas 

milg15:
Allows to drag & drop an image into the canvas

| Drag | Drop |
| ------------- | ------------- |
|  | 

* Clean up of (#156)

### DIFF
--- a/packages/app/src/containers/Canvas.tsx
+++ b/packages/app/src/containers/Canvas.tsx
@@ -60,7 +60,12 @@ function getCursorForDirection(
   }
 }
 
-function getPoint(event: MouseEvent): Point {
+export type OffsetPoint = {
+  offsetX: number;
+  offsetY: number;
+};
+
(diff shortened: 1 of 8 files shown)


packages/app/src/containers/Canvas.tsx
```diff
@@ -734,26 +735,57 @@ export default memo(function Canvas() {
     }
   }, [state, handleDirection]);
 
+  const onDropFile = useCallback(
+    async (file: File, extension: string, e: MouseEvent) => {
```
dabbott, line 739: Since we only use the event for getting the drop location, we could pass just a point instead of the event object, and update our `getPoint` function to use the same type.

```ts
export type OffsetPoint = {
  offsetX: number;
  offsetY: number;
};

function getPoint(event: OffsetPoint): Point { ... }
```

dabbott (approved):

PR #169, opened 2021-07-14
Degenerate paths and update line behavior

### LINKED ISSUE #157: Degenerate paths gracefully
When a horizontal/vertical line have the same y value in the inspector, the path shouldn't disappear or show NaN.

leannepepper:
This PR will address issues with editing and interacting with lines, defined as any layer containing 2 straight points. 
This includes: 
1) Coordinate inspector will show both start and end points of a line 
2) Update to drag handles
3) Enforce a height of 1 on line bounding rects
 closes #157


packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -114,6 +121,42 @@ export function pointReducer(
         );
       });
     }
+
+    case 'setLinePointX':
+    case 'setLinePointY': {
```
dabbott, line 126: There might be an opportunity to combine these with `setPointX`/`setPointY` by passing a `SelectedPointLists`. I'm not sure if it'd be an improvement... but I suspect it would

packages/noya-state/src/selection.ts
```diff
@@ -16,10 +18,11 @@ const compassDirectionMap: Record<CompassDirection, Point> = {
   w: { x: 0, y: 0.5 },
   nw: { x: 0, y: 0 },
 };
+export const dragHandelSize: number = 7;
```
dabbott, line 21: `Handle`

packages/noya-state/src/selection.ts
```diff
@@ -41,3 +44,32 @@ export function getDragHandles(
...
+        compassDirection === 'w' ? startDecodedPoint : endDecodedPoint;
+
+      return {
+        rect: {
+          x: directionPoint.point.x - handleSize / 2,
+          y: directionPoint.point.y - handleSize / 2,
+          width: handleSize,
+          height: handleSize,
+        },
+        compassDirection,
```
dabbott, line 72: If it's not too hard, we might want to pick the direction based on angle like Sketch. If it's hard, this is fine for now (I will have some other small comments if we keep this)

packages/noya-state/src/selectors/pointSelectors.ts
```diff
@@ -233,6 +240,14 @@ export const moveSelectedPoints = (
         y: getNewValue(layer.frame.y, mode, delta.y),
       };
 
+      if (draftLayer.frame.height === 0) {
+        draftLayer.frame.height = 1;
```
dabbott, line 244: When we adjust the layer like this, do we need to move the points to `y = 0.5`? Otherwise they'll presumably be at the top or bottom of the frame, which isn't quite right

packages/noya-state/src/selection.ts
leannepepper: I added some code to calculate the angle between start and end points, however I'm having a hard time finding the relationship between the angle and the direction.

packages/noya-state/src/selectors/pointSelectors.ts
```diff
@@ -233,6 +260,26 @@ export const moveSelectedPoints = (
         y: getNewValue(layer.frame.y, mode, delta.y),
       };
 
+      // if (draftLayer.frame.height === 0) {
```
leannepepper, line 263: Do we still need this?

packages/noya-state/src/selection.ts
leannepepper: Also would we use this angle or the rotation on layer.rotation?

packages/app/src/components/inspector/LineInspector.tsx
```diff
@@ -0,0 +1,189 @@
...
+    () => onSetIsFlippedVertical(!isFlippedVertical),
+    [isFlippedVertical, onSetIsFlippedVertical],
+  );
+
+  const handleSetIsFlippedHorizontal = useCallback(
+    () => onSetIsFlippedHorizontal(!isFlippedHorizontal),
+    [isFlippedHorizontal, onSetIsFlippedHorizontal],
+  );
+
+  const flipButtonElements = useMemo(
```
dabbott, line 76: We should make a new component for these flip controls so we don't have to duplicate them

packages/app/src/containers/Inspector.tsx
```diff
@@ -108,6 +109,9 @@ export default memo(function Inspector() {
       Layers.isBitmapLayer(l),
     );
     const hasTextLayer = selectedLayers.some((l) => Layers.isTextLayer(l));
+    const hasLineLayer = selectedLayers.some(
```
dabbott, line 112: We can allow editing multiple lines at once. It looks like Sketch show the LineInspector when every selected layer is a line. We can do this by changing `.some` to `.every`

packages/noya-state/src/reducers/pointReducer.ts
```diff
@@ -14,7 +15,12 @@ import { ApplicationState, SetNumberMode } from './applicationReducer';
 export type PointAction =
   | [type: 'setPointCurveMode', curveMode: Sketch.CurveMode]
   | [type: 'setPointCornerRadius', amount: number, mode?: SetNumberMode]
-  | [type: 'setPointX' | 'setPointY', amount: number, mode?: SetNumberMode]
+  | [
+      type: 'setPointX' | 'setPointY',
+      amount: number,
+      mode?: SetNumberMode,
+      direction?: 'start' | 'end',
```
dabbott, line 22: Maybe this would be simpler if we pass in the `SelectedPointLists`? We can make it a required argument, and move it to position 2 in the tuple so the API looks more consistent with others that take ids.

```ts
[
      type: 'setPointX' | 'setPointY',
      pointLists: SelectedPointLists,
      amount: number,
      mode?: SetNumberMode,
]
```

packages/noya-renderer/src/components/DragHandles.tsx
```diff
@@ -13,9 +23,30 @@ interface Props {
 
 export default memo(function DragHandles({ selectionPaint, rect }: Props) {
   const CanvasKit = useCanvasKit();
+  const [state] = useApplicationState();
 
   const dragHandlePaint = useColorFill(CanvasKit.Color(255, 255, 255, 1));
-  const dragHandles = getDragHandles(rect);
+
+  let dragHandles = getDragHandles(rect);
```
dabbott, line 30: Maybe rename `getDragHandles` => `getRectDragHandles` for consistency with the line version

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -62,13 +63,32 @@ const BoundingRect = memo(function BoundingRect({
...
+  const [state] = useApplicationState();
 
   const alignedRect = useMemo(
     () => Primitives.rect(CanvasKit, insetRect(rect, 0.5, 0.5)),
     [CanvasKit, rect],
   );
 
-  return <RCKRect rect={alignedRect} paint={selectionPaint} />;
+  let isShapeALine = false;
+  if (state.selectedObjects.length === 1) {
```
dabbott, line 74: I'd recommend moving this stuff out of `BoundingRect`. If we don't need a bounding rect, we can short-circuit before we render `<BoundingRect />`.

We can use the `getSelectedLineLayer` selector I suggest in my other comment

packages/noya-renderer/src/components/DragHandles.tsx
```diff
@@ -13,9 +23,30 @@ interface Props {
...
+
+  let isShapeALine = false;
```
dabbott, line 32: Can we do a selector function for this? Roughly:

```ts
const lineLayer = Selectors.getSelectedLineLayer(state) 
const dragHandles = lineLayer 
  ? getLineDragHandles(lineLyaer.frame, lineLayer.points) 
  : getRectDragHandles()
```

packages/noya-renderer/src/components/DragHandles.tsx
```diff
@@ -13,9 +23,30 @@ interface Props {
...
+  if (state.selectedObjects.length === 1) {
+    const page = Selectors.getCurrentPage(state);
+    const indexPath = Layers.findIndexPath(
+      page,
+      (layer) => layer.do_objectID === state.selectedObjects[0],
+    );
+
+    if (indexPath) {
+      const layer = Layers.access(page, indexPath);
+      isShapeALine = isPointsLayer(layer) ? isLine(layer.points) : false;
```
dabbott, line 43: This boolean logic can be simplified when we move it into a selector

packages/noya-state/src/selection.ts
dabbott: I'm not sure, but let's get the drag handles rendering in the right positions, cause otherwise I think it'll be hard to tell if this is working or not

packages/noya-state/src/selectors/pointSelectors.ts
dabbott: I don't think so, not till we generalize to _any_ degenerate path instead of just a straight line with 2 points. We can remove it for now and come back to it

dabbott (changes requested):

packages/noya-renderer/src/components/SketchFileRenderer.tsx
leannepepper: created `omitBoundingRect` function

dabbott (approved):

PR #176, opened 2021-07-17
Add to artboard/page on drag 

milg15:


packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -264,13 +274,77 @@ export function canvasReducer(
...
+
+        if (
+          layerInfo.every((l) => l.parent === parent) ||
+          layerInfo.some(
+            (l) => Layers.isArtboard(l.layer) || Layers.isSymbolMaster(l.layer),
+          )
+        )
+          return;
+
+        deleteLayers(indexPaths, draftPage);
```
dabbott, line 311: If we delete layers, we won't be able to use the `parentIndexPath` anymore, since that will potentially change the layer hierarchy.

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -264,13 +274,77 @@ export function canvasReducer(
...
+      const indexPaths = getSelectedLayerIndexPathsExcludingDescendants(state);
+
+      const parentIndexPath = Layers.findIndexPath(
+        state.sketch.pages[pageIndex],
+        (layer) =>
+          (Layers.isArtboard(layer) || Layers.isSymbolMaster(layer)) &&
+          rectContainsPoint(layer.frame, point),
+      );
+
+      return produce(state, (draft) => {
```
dabbott, line 290: After thinking about it some more, I think we should try to re-use `'moveLayer'` here. This code is kinda tricky, so it's better if we only have it in one place. Let's make a new function in `layerReducer`:

```ts
export function moveLayer(
  state: ApplicationState,
  id: string | string[],
  destinationId: string,
  rawPosition: RelativeDropPosition,
) {
  // ...
  return produce(state, draft => { ... })
}
```

And move the [entire 'moveLayer' case](https://github.com/noya-app/noya/blob/f60c0ef1af834765386ea9fc649aa3b73e8f7a8b/packages/noya-state/src/reducers/layerReducer.ts#L197-L285) into that function.

Then, here within `'moveLayersIntoParentAtPoint'`, we can call it:

```ts
const parentId = mouseIsInParentLayer(page, point)?.do_objectID ?? page.do_objectID

return moveLayer(state, state.selectedObjects, parentId, 'inside');
```

If there's any behavior that's special to this dragging use case, we could make it another param of `moveLayer()`, or handle it before calling `moveLayer()`

dabbott (changes requested):
At first I thought this worked perfectly, and then I realized it sometimes crashes depending on the layer hierarchy. I left a comment about why, but I think it'll be safer to re-use the `moveLayer` code that (as far as I know) already handles this safely

packages/noya-state/src/reducers/canvasReducer.ts
milg15: the move layer function has a "position" variable that I don't need in the other drag function, and it's used right in the middle of the moveLayer function. Do I remove it ?🤔

milg15: Oh I see how it's supposed to work

milg15: okey I see it now 😆 I think all good

dabbott (approved):
Awesome, looks great!

PR #182, opened 2021-07-20
Modify Gradient from Canvas

milg15:


dabbott:
TODO:

- Correctly handling gradient selection state when changing `fillType` in the fill popover
- Canvas: Add gradient stop by clicking the line on the canvas (I commented this out)
  - This should be a new action, like `['addStopToGradient', Point]`, and it can use `state.selectedGradient` within the reducer. I don't think it needs to go through `interactionState` at all.
- Gradient picker: Inserting new gradient stops in the correct position, or insert and sort
- Border gradients

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -141,6 +147,65 @@ export function canvasReducer(
...
+
+        const color = rgbaToSketchColor(interpolateRgba(gradient, position));
+        gradientStops.push({
+          _class: 'gradientStop',
+          color,
+          position,
+        });
+
+        gradientStops.sort((a, b) => a.position - b.position);
+        const nexIndex = gradientStops.findIndex(
```
dabbott, line 200: `nextIndex`

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -141,6 +147,65 @@ export function canvasReducer(
...
+
+        if (!gradientStops) return;
+
+        const gradient = gradientStops.map((g) => ({
+          color: sketchColorToRgba(g.color),
+          position: g.position,
+        }));
+
+        const color = rgbaToSketchColor(interpolateRgba(gradient, position));
+        gradientStops.push({
```
dabbott, line 193: Very minor: we can use `SketchModel.gradientStop({ color, position })` to create a gradient stop, which automatically adds the `_class` property

packages/noya-state/src/reducers/applicationReducer.ts
```diff
@@ -51,6 +53,13 @@ export type SelectedControlPoint = {
   controlPointType: controlPointType;
 };
 
+export type SelectedGradient = {
+  layerId: string;
+  fillIndex: number;
+  stopIndex: number;
+  styleType: 'fill' | 'border';
```
dabbott, line 60: I just realized, if we change this to `'fills' | 'borders'` we, can simplify some of our other code by using the `styleType` directly to look up the correct property, e.g.:

```ts
if (layer.style?.[styleType]?.[fillIndex].fillType !== Sketch.FillType.Gradient) { return state }

const gradientStops = layer.style?.[styleType]?.[fillIndex].gradient.stops;
```

packages/noya-state/src/selectors/gradientSelectors.ts
```diff
@@ -0,0 +1,203 @@
...
+  );
+
+  if (!selectedLayerGradientPoints) return -1;
+
+  return selectedLayerGradientPoints.findIndex((gradientPoint) =>
+    isPointInRange(gradientPoint.point, point, 5.5),
+  );
+}
+
+function isPointOnLine(A: Point, B: Point, point: Point) {
```
dabbott, line 169: Let's move this into `line.ts` in `noya-geometry` and add a couple tests for it.

dabbott: Is this equivalent? If so I think it's a little clearer

```ts
return Math.abs(d1 + d2 - lineLength) < 0.5
```

dabbott: We should change the argument types to `function isPointOnLine(point: Point, line: [Point, Point])` to be a little clearer and consistent with `getClosestPointOnLine`. And rename `lineLen` => `lineLength`, since we don't usually use abbreviations

packages/noya-state/src/selectors/gradientSelectors.ts
```diff
@@ -0,0 +1,203 @@
...
+): number {
+  const selectedLayerGradientPoints = getSelectedGradientStopPoints(
+    state,
+    true,
+  );
+
+  if (!selectedLayerGradientPoints) return -1;
+
+  return selectedLayerGradientPoints.findIndex((gradientPoint) =>
+    isPointInRange(gradientPoint.point, point, 5.5),
```
dabbott, line 165: I'm guessing this 5.5 is to match the `POINT_RADIUS * 1.5` I added? Maybe we should make a constant like `SELECTED_GRADIENT_POINT_RADIUS` or something for both of these.

dabbott (changes requested):
This looks amazing 💯

It seems like sometimes I'm not able to drag border gradient points. I did see it work once, but most of the time I'm only able to select and add new points. It always works with fills. Does that happen for you? If not, I can debug after we get the other changes in

packages/noya-designsystem/src/components/GradientPicker.tsx
```diff
@@ -41,6 +34,18 @@ export default memo(function GradientPicker({
   onDelete,
   onSelectStop,
 }: Props) {
+  const colorModel: ColorModel<RgbaColor> = {
```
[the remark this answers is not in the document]
milg15: Ohh! there is a weird rendering error. When I have it outside. I am not sure on why actually. I forgot

This is the error `TypeError: colorModel.toHsva is not a function`

packages/noya-state/src/selectors/gradientSelectors.ts
milg15: How so? Should I still call it from the function but with constant instead of the direct value?

packages/noya-designsystem/src/components/GradientPicker.tsx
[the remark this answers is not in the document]
dabbott: Ok! Let's leave it here but wrap it in a `useMemo`

dabbott (approved):

PR #207, opened 2021-08-06
More gradient actions

milg15:
- [ ]  Handle when the popup is still open but you change layer. 
- [x] Fix some sorting issue when 360 some points


milg15:
A new user left a comment. This user must be approved by a Netlify team owner before comments can be displayed.

[Approve this user](https://app.netlify.com/teams/dabbott/reviewers)

packages/noya-state/src/selectors/gradientSelectors.ts
```diff
@@ -177,3 +178,39 @@ export function isPointerOnGradientLine(state: ApplicationState, point: Point) {
...
+
+  const center = selectedLayerGradientPoints[0].point;
+  const lastPoint =
+    selectedLayerGradientPoints[selectedLayerGradientPoints.length - 1].point;
+
+  const height = distance(center, lastPoint) * 2;
+  const width = height * selectedGradient.elipseLength;
+
+  // Maybe there is a function to do this in a more efficient way
+  const theta =
```
dabbott, line 203: We can use an `AffineTransform` here:

```ts
  const position = AffineTransform.rotate(theta, center.x, center.y).applyTo({
    x: center.x - width / 2,
    y: center.y,
  });
```

We should try to share this code with the `SketchFileRenderer`. Maybe:

```ts
function getEllipseEditorPoint(...): Point { ... }
```

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -593,6 +636,46 @@ export function canvasReducer(
...
+            )
+              return;
+            const gradient = draftLayer.style?.fills?.[fillIndex]?.gradient;
+            const points = getSelectedGradientStopPoints(state);
+            if (!points) return;
+            const center = points[0].point;
+            const lastPoint = points[points.length - 1].point;
+
+            const radius = distance(center, lastPoint);
+            const length = distance(center, { x: current.x, y: center.y });
```
dabbott, line 674: At larger distances, the movement jumps right at the start of the drag, and is slightly different than the mouse position.

I think this should fix it:

```ts
const length = distance(current, getClosestPointOnLine(current, [center, lastPoint]))
```

packages/noya-state/src/reducers/canvasReducer.ts
```diff
@@ -593,6 +636,46 @@ export function canvasReducer(
 
             break;
           }
+          case 'maybeMoveGradientElipseLength': {
```
dabbott, line 639: In general the `maybe` states do 2 things:

1. Let us take a snapshot of the page data so we can always do calculations with the state at the start of the interaction. This is important since sometimes interactions can temporarily have an invalid state (e.g. scaling a layer to a width of 0)
2. Enforce a dragging threshold of a few pixels, so the user doesn't accidentally adjust things by a tiny amount

We don't really need (1) in this case, since there's no "invalid" ellipse length. However it's still useful for (2), so I think we should keep it.

We can delete this empty switch statement though.

dabbott (changes requested):
Creating a new radial gradient is kinda awkward, cause the ellipse and gradient line are exactly the same.

Screen Shot 2021-08-06 at 9 21 54 AM

The easiest solution is to change the default `elipseLength` from 0 to 1. We can do this for all gradients here: https://github.com/noya-app/noya/blob/a57e4d16ee1b0841153be68dd068000e61e6281e/packages/noya-sketch-model/src/index.ts#L72

Whenever it's our own code, we should spell "ellipse" correctly with 2 "L", rather than Sketch's incorrect "elipse". So we should update all our variable/action names.

dabbott:
Creating a new radial gradient is kinda awkward, cause the ellipse and gradient line are exactly the same.

Screen Shot 2021-08-06 at 9 21 54 AM

Originally I was thinking we could set the default `elipseLength` to 1 to solve this. But, layers created from Sketch have a default of 0. It looks like Sketch handles 0 as a special case, and displays the gradient and editor UI as if it were 1

packages/noya-state/src/selectors/gradientSelectors.ts
milg15: 🤔 I use theta and those points in different ways. The position of the drawn rectangle and the "caption" rectangle are a bit different.

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -350,8 +353,66 @@ export default memo(function SketchFileRenderer() {
...
+      getCurrentPage(state),
+      state.selectedGradient,
+    );
+
+    if (!selectedGradient) return null;
+
     const { stopIndex } = state.selectedGradient;
 
+    let radialGradient = null;
+    if (selectedGradient.gradientType === Sketch.GradientType.Radial) {
```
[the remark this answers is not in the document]
milg15: I think I can create one file that also handles Linear and Angel, all three since they all use the same information but show it in different ways.

packages/noya-renderer/src/components/SketchFileRenderer.tsx
```diff
@@ -439,7 +378,7 @@ export default memo(function SketchFileRenderer() {
             {(state.selectedObjects.length > 1 ||
               !Selectors.getSelectedLineLayer(state)) &&
               boundingRect &&
-              !gradientPoints &&
+              !gradientStopPoints &&
```
dabbott, line 381: Better to use `!state.selectedGradient` here and below

packages/noya-renderer/src/components/GradientEditor.tsx
```diff
@@ -0,0 +1,152 @@
...
+        />
+      )}
+      {angularGradient}
+      {radialGradient}
+      {gradientStopPoints.map(({ point, color }, index) => {
+        const path = new CanvasKit.Path();
+
+        const radius =
+          index === stopIndex
+            ? Selectors.POINT_RADIUS * 1.5
```
dabbott, line 128: I think we already made a constant that we can use for this instead of 1.5?

dabbott (changes requested):

packages/noya-state/src/selectors/gradientSelectors.ts
```diff
@@ -177,3 +237,89 @@ export function isPointerOnGradientLine(state: ApplicationState, point: Point) {
...
+    page,
+    (layer) => layer.do_objectID === layerId,
+  );
+
+  if (!indexPath) return;
+
+  const layer = Layers.access(page, indexPath);
+  const bounds = createBounds(layer.frame);
+
+  const horizontalDistance = distance(
```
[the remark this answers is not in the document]
milg15: If we use max doesn't that make the circle be outside of the frame?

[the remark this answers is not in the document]
dabbott: Yeah, that's how Sketch displays it 🤷‍♂️

Screen Shot 2021-08-11 at 5 24 06 PM
Screen Shot 2021-08-11 at 5 24 24 PM

packages/noya-state/src/selectors/gradientSelectors.ts
```diff
@@ -77,24 +82,40 @@ export function getSelectedGradientStopPoints(
...
+  const stops = sorted
+    ? [...gradient.stops].sort((a, b) => a.position - b.position)
+    : gradient.stops;
+
+  if (gradient.gradientType === Sketch.GradientType.Angular) {
+    const circumference = getAngularGradientCircumference(state);
+    if (!circumference) return;
+
+    return stops.map((stop) => {
+      const radians = stop.position * Math.PI * 2;
```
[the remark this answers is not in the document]
milg15: toRadiands turn degrees into radians but we aren't using a degree but a percentage

packages/noya-renderer/src/components/GradientEditor.tsx
milg15: We do but they have different values. The selection radius for clicking and the one for showing it are different sizes 🤔 The one that it's used when clicking it's a bit bigger than the one we show

packages/noya-renderer/src/components/GradientEditor.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+    const height = len * 2;
+    const width =
+      gradient.elipseLength === 0 ? height : height * gradient.elipseLength;
+
+    const x = center.x - width / 2;
+    const y = center.y - len;
+
+    const theta =
+      Math.atan2(point.y - center.y, point.x - center.x) - Math.PI / 2;
```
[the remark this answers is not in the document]
milg15: sort of, but the one in the function it's being rotated so I can't call it here but maybe I can create a better function that can work in both places where I use this information. 🤔

packages/noya-geometry/src/line.ts
```diff
@@ -42,3 +43,22 @@ export function isPointInLine(point: Point, line: [Point, Point]) {
 
   return Math.abs(d1 + d2 - lineLength) < 0.5;
 }
+
+// TODO: Not really accurate, but works for now.
+export function isPointInCircle(
```
dabbott, line 48: We should rename this `isPointOnCircumference`, since the current name implies that any point within the circle would give `true`

packages/noya-state/src/selectors/gradientSelectors.ts
```diff
@@ -177,3 +236,86 @@ export function isPointerOnGradientLine(state: ApplicationState, point: Point) {
...
+  const { layerId } = state.selectedGradient;
+  const page = getCurrentPage(state);
+  const indexPath = Layers.findIndexPath(
+    page,
+    (layer) => layer.do_objectID === layerId,
+  );
+
+  if (!indexPath) return;
+
+  const parent = getParentLayer(page, indexPath);
```
dabbott, line 304: We need to consider every ancestor here, e.g.

```ts
const layer = Layers.access(page, indexPath)
const transform = getLayerTransformAtIndexPath(page, indexPath)
    .transform(getLayerFlipTransform(layer))
    .transform(getLayerRotationTransform(layer))
const bounds = createBounds(layer.frame)
const center = transform.applyTo({ x: bounds.midX, y: bounds.midY })
```

Otherwise, the same problem will happen if there's more than one parent:

Screen Shot 2021-08-12 at 10 54 49 AM

We should also handle rotations. I think the way to do that with the fewest changes would be to return a `rotation` value from this function.

```ts
const line = transform.applyTo({ x: bounds.maxX, y: bounds.midY })
const rotation = Math.atan2(line.y - center.y, line.x - center.x)

return { center, radius, rotation }
```

Then we can use the rotation to position the gradient stops, e.g. in `getSelectedGradientStopPoints`:

```ts
const radians = stop.position * Math.PI * 2 + circle.rotation
```

Right now, we're not handling rotation and flip, so the stops will appear in the wrong place:

Screen Shot 2021-08-12 at 11 41 32 AM

dabbott (changes requested):

dabbott (approved):

PR #223, opened 2021-08-17
Gradient Test

milg15:


packages/noya-state/src/selectors/__tests__/__snapshots__/gradientSelectors.test.ts.snap
```diff
@@ -0,0 +1,65 @@
...
+    },
+    "point": Object {
+      "x": 50,
+      "y": 100,
+    },
+  },
+]
+`;
+
+exports[`get Gradient Point sorted points 1`] = `undefined`;
```
dabbott, line 65: Should this one have values? We should have an unsorted gradient as input, and double check that it's sorted here.

packages/noya-state/src/selectors/__tests__/__snapshots__/gradientSelectors.test.ts.snap
```diff
@@ -0,0 +1,65 @@
...
+      "blue": 1,
+      "green": 1,
+      "red": 1,
+    },
+    "point": Object {
+      "x": 100,
+      "y": 50,
+    },
+  },
+  Object {
```
dabbott, line 18: We should add another point in the middle of our gradient, since `from`/`to` are sort of special cases that may not use `lerp`

packages/noya-state/src/selectors/__tests__/gradientSelectors.test.ts
```diff
@@ -0,0 +1,217 @@
...
+    SketchModel.fill({
+      fillType: Sketch.FillType.Gradient,
+      gradient: SketchModel.gradient({
+        gradientType: Sketch.GradientType.Angular,
+      }),
+    }),
+  ],
+});
+
+const linearGradienStyle = SketchModel.style({
```
dabbott, line 21: linearGradientStyle

packages/noya-state/src/selectors/__tests__/gradientSelectors.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      styleType: 'fills',
+    };
+
+    expect(getAngularGradientCircle(state)).toEqual({
+      center: { x: 50, y: 50 },
+      radius: 50,
+      rotation: 0,
+    });
+  });
+  test('fliped layer', () => {
```
dabbott, line 59: flipped

dabbott: We should also test `rotation`, which is different than flipping. And maybe `isFlippedVertical` too?

packages/noya-state/src/selectors/__tests__/gradientSelectors.test.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    expect(getAngularGradientCircle(state)).toEqual({
+      center: { x: 550, y: 550 },
+      radius: 50,
+      rotation: 0,
+    });
+  });
+});
+
+describe('get Gradient Point', () => {
```
dabbott, line 150: We should test this function on layers within an artboard/group too

packages/noya-state/src/selectors/__tests__/layerSelectors.test.ts
```diff
@@ -1,7 +1,9 @@
 import produce from 'immer';
 import { debugDescription, SketchModel } from 'noya-sketch-model';
+
 import { fixGroupFrame, resizeLayerFrame } from '../layerSelectors';
 
+// Layer Selectors
```
dabbott, line 6: I don't think we need the comment, since we already the test name

dabbott (changes requested):

packages/noya-state/src/selectors/__tests__/gradientSelectors.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      layerId: rectangle.do_objectID,
+      fillIndex: 0,
+      stopIndex: 0,
+      styleType: 'fills',
+    };
+
+    expect(getAngularGradientCircle(state)).toEqual({
+      center: { x: 50, y: 50 },
+      radius: 50,
+      rotation: Math.PI.valueOf(),
```
dabbott, line 75: Do we need `valueOf`?

milg15: It gave me an error at first but it seems to be working now

dabbott (approved):

PR #273, opened 2021-09-03
Copy and Paste Layers

milg15:


milg15:
Should we support pasting directly into the canvas in this issue?

packages/noya-state/src/selectors/layerSelectors.ts
```diff
@@ -70,6 +70,21 @@ export const getSelectedLayers = (
...
+  state: Draft<ApplicationState>,
+): PageLayer[] => {
+  const page = getCurrentPage(state);
+
+  const selectedLayers = Layers.findAll(page, (layer) =>
+    state.selectedLayerIds.includes(layer.do_objectID),
+  ) as PageLayer[];
+
+  return selectedLayers.map((layer) => {
+    resizeLayerFrame(layer, page.frame);
```
dabbott, line 83: Does this work for nested layers? What about rotated layers? I would think we'd need to do this using transforms, but if this works then it's fine.

Since `resizeLayerFrame` mutates the layer object, we would probably want to wrap things in a `produce` to first clone the layers.

dabbott (changes requested):
Nice! I'm glad we were able to find a way to make copy & paste work normally

packages/noya-state/src/selectors/layerSelectors.ts
milg15: ohh I think not

packages/app/src/hooks/useLayerMenu.tsx
```diff
@@ -118,6 +120,8 @@ export default function useLayerMenu(
 
   const canShow = layers.some((item) => !item.isVisible);
 
+  const canPaste = true;
```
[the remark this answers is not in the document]
milg15: I am not 100% sure what layers can be pasted 🤔

packages/noya-state/src/selectors/layerSelectors.ts
milg15: I don't even remember why I was using the `resizeLayerFrame` 😆 I just noticed it's not even been called.

milg15: Okey it does work for groups 😆 It works. I only need to delete the resize layer that I don't even need. I think this how I was trying to do it before and left it like this. I just need the selected layer

packages/app/src/hooks/useLayerMenu.tsx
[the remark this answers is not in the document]
milg15: This was for the menu but the menu 'Paste' opinion doesn't really work in Chrome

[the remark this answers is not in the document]
milg15: I was trying to check but I don't think we can do it. I added a small code it's commented you can tested it but I'm not sure how to do it synchronously.

PR #275, opened 2021-09-06
Update text style of new text

milg15:


packages/noya-state/src/reducers/textStyleReducer.ts
```diff
@@ -178,13 +201,23 @@ export function textStyleReducer(
...
-            (layer) => {
+            (layer, index) => {
               if (!Layers.isTextLayer(layer) || !Layers.hasTextStyle(layer))
                 return;
 
               const attributes = layer.style.textStyle.encodedAttributes;
 
               attributes.MSAttributedStringTextTransformAttribute = action[1];
+
+              if (index === 0) {
```
dabbott, line 212: We do this check for `index === 0` in a few places, and I think it should work, but it's a little unclear what it's for. I think it might be clearer if we do that mutation outside the loop. The also solves the potential issue of e.g. a rectangle as the first selected layer, which would cause the `index === 0` check to not work.

```ts
return produce(state, (draft) => {
   const textLayers = accessPageLayers(draft, pageIndex, layerIndexPaths).filter(Layers.isTextLayer)
  
   textLayers.forEach(...)

   const firstTextLayer = textLayers[0]

   if (firstTextLayer) {
      draft.lastEditedTextStyle = ...
   }
 })
```

dabbott (changes requested):
I just tried it, and this looks & works great!

dabbott:
@milg15 Can we do the same change for the 2 other places where we have an `index === 0` check?

milg15:
> index === 0
Ohh! I didn't notice

dabbott (approved):
Nice, looks great! 💯

PR #278, opened 2021-09-12
finalize site

souporserious:
Adds most of the final touches to the site. There are some parts still ripe for cleanup in the index file. Some things can probably be further pulled out into components and tokens that I can do in a follow-up PR as we shape the compiler.


packages/site/src/pages/index.tsx
```diff
@@ -1,11 +1,26 @@
...
+import { Packages } from '../components/Packages';
+
 import logoSrc from '../assets/logo.svg';
-import * as theme from '../theme';
+import sketchSrc from '../assets/sketch.svg';
+import figmaSrc from '../assets/figma.svg';
+import xdSrc from '../assets/xd.svg';
 
-const preventTextOprhan = (text: string) => text.replace(/ /g, '\u00a0');
+const preventTextRunt = (text: string) => text.replace(/ /g, '\u00a0');
```
souporserious, line 23: Was using the wrong terminology 😇  more info [here](https://brennenreece.squarespace.com/blog/fixing-runts-in-indesign-using-grep) if you're curious.

packages/site/src/pages/index.tsx
```diff
@@ -36,148 +51,418 @@ const features = [
...
+            specific needs.
+          </Text>
         </Stack>
 
-        <Spacer.Vertical size="12rem" />
+        <Spacer
+          size="6rem"
+          css={{
+            [mediaQuery.medium]: {
+              minHeight: '24rem',
```
souporserious, line 206: Not happy with this and can cleanup later, instead of `size` maybe just `minHeight` for now. I was only using `Spacer` for vertical alignment since columns should be used for horizontal alignment.

packages/site/src/pages/index.tsx
```diff
@@ -36,148 +51,418 @@ const features = [
...
+        css={{
+          minHeight: '6rem',
+          [mediaQuery.medium]: {
+            minHeight: '8rem',
+          },
+        }}
+      />
+
+      <Grid
+        gridColumn="1 / -1"
```
souporserious, line 86: Style props fall apart a little here with responsive. We could do something like an object:
```jsx
<Grid column={{ initial: '1 / -1', medium: '2 / span 10'  }} />
```
But I didn't have time to spend on that. This was something I solved recently in JSXUI I'd love to discuss further.

packages/site/src/components/PagerView.tsx
```diff
@@ -44,21 +53,29 @@ export function View({ color }: { color: string }) {
   };
 
   return (
+    // @ts-ignore
```
souporserious, line 56: There was a weird circular reference error being thrown here that I couldn't figure out the cause. Possibly the `css` prop type conflicting with react-spring.

dabbott:
Merging so this doesn't get out of sync

PR #280, opened 2021-09-26
Toggle Lock

milg15:
#279


dabbott:
This is clever, but I think it's gonna be problematic for certain actions, like scaling 🤔. We do still want the layers to be measured even when they're locked.

https://user-images.githubusercontent.com/1198882/141029337-f6bce957-9cbb-43ce-ae69-bb8ad0916687.mov

We might need to handle locking in the individual actions.

dabbott:
To start I made it so scaling is locked.

milg15:
Oh what's missing in this one?

dabbott:
Scaling being locked is a good start. Possibly preventing a drag when 2 layers selected from moving the locked layer... but Sketch actually allows this, so we can leave it 🤷‍♂️

PR #283, opened 2021-10-11
scroll  bar in symbols

milg15:


packages/noya-designsystem/src/components/ContextMenu.tsx
```diff
@@ -149,6 +149,7 @@ export interface MenuProps<T extends string> {
   items: MenuItem<T>[];
   onSelect: (value: T) => void;
   isNested?: boolean;
+  isScroll?: boolean;
```
dabbott, line 152: Let's call this prop `scrollable`, since we have that in a few other places

packages/noya-designsystem/src/components/DropdownMenu.tsx
```diff
@@ -148,6 +152,7 @@ function DropdownMenuRoot<T extends string>({
         </RadixDropdownMenu.Trigger>
       )}
       <RootElement
+        style={style}
```
dabbott, line 155: Instead of passing a style here, let's pass the `scrollable` prop into this component, and handle a `scrollable` prop here in the `contentStyle` function that's shared by both context menus and dropdown menus:

https://github.com/noya-app/noya/blob/9cfbf15329ab6dd103a3a3429de5a183d07337d0/packages/noya-designsystem/src/components/internal/Menu.tsx#L82

We pass `contentStyle` when setting up the `RootElement`: https://github.com/noya-app/noya/blob/9cfbf15329ab6dd103a3a3429de5a183d07337d0/packages/noya-designsystem/src/components/DropdownMenu.tsx#L116

packages/noya-designsystem/src/components/DropdownMenu.tsx
```diff
@@ -134,6 +135,9 @@ function DropdownMenuRoot<T extends string>({
     [isNested, items, onSelect, shouldBindKeyboardShortcuts],
   );
 
+  const style = isScroll
```
dabbott, line 138: I like this maxHeight approach cause it's simple, and we could make the maxHeight a prop. 

I think it'd be ideal if the menu extended to the bottom of the screen, minus some padding. How difficult do you think that'd be? It it's hard it might not be worth it.

dabbott (changes requested):

packages/noya-designsystem/src/components/internal/Menu.tsx
```diff
@@ -79,13 +79,24 @@ export const styles = {
...
+  }): CSSObject => ({
     borderRadius: 4,
     backgroundColor: theme.colors.popover.background,
     color: theme.colors.text,
     boxShadow: '0 2px 4px rgba(0,0,0,0.2), 0 0 12px rgba(0,0,0,0.1)',
     padding: '4px',
     border: `1px solid ${theme.colors.divider}`,
+    ...(scrollable && {
+      height: '100%',
+      maxHeight: 'calc(100vh - 80px)',
```
dabbott, line 97: I think this will only work because the dropdown is at the top of the screen, but we can leave it for now, since that's the only place we're using this

PR #286, opened 2021-11-12
unit test Layer Reducer

milg15:


packages/noya-state/src/reducers/__tests__/layerReducer.test.ts
```diff
@@ -530,3 +530,93 @@ describe('grouping', () => {
...
+
+    const updated = layerReducer(
+      groupedState,
+      ['addLayer', rectangle],
+      context,
+    );
+
+    expect(
+      debugDescription([
+        Selectors.getCurrentPage(state),
```
dabbott, line 554: Let's use groupedState here, since that will show a clearer before/after

packages/noya-state/src/reducers/__tests__/layerReducer.test.ts
```diff
@@ -530,3 +530,93 @@ describe('grouping', () => {
...
+        Selectors.getCurrentPage(updated),
+      ]),
+    ).toMatchSnapshot();
+  });
+  test('original layer is in parent layer', () => {
```
dabbott, line 559: What do we do differently if the original layer is in the parent? Does the parent need to be selected? I'm not sure I understand this one

dabbott (changes requested):
This looks great and helps a lot!

I think as a next PR, we should make the update: if the original layer is within view, the pasted layer should be placed at the same position, rather than in the center of the canvas

We should be able to see that change in the snapshots, which will be great

packages/noya-state/src/reducers/__tests__/layerReducer.test.ts
milg15: Depends on which layer it's selected. 
1. If a selected layer it's a parent layer but not a page (group, artboard, etc) the pasted element would be added to that layer at the end. 
2. If the parent layer it's a page and the original layer it's on the page we'll add it be under that element
3. If the original layer is not on the page and we have a layer selected, it'll add it under that layer.
4. If we don't have any layer selected just the page it will just add it at the end

I hope that makes sense 😅

dabbott: Got it! I'll have to play with it some more to see if that feels right, but very helpful to have the tests to demonstrate that those different cases are working correctly

dabbott (approved):

PR #288, opened 2021-11-14
Update on the paste layers

milg15:
I can delete the comments later 😅

### DIFF
--- a/packages/noya-state/src/reducers/layerReducer.ts
+++ b/packages/noya-state/src/reducers/layerReducer.ts
@@ -30,12 +30,16 @@ import {
   LayerIndexPaths,
   moveLayer,
   removeLayer,
-  isLayerVisible,
+  getLayersInRect,
 } from '../selectors/selectors';
 import { SelectionType, updateSelection } from '../utils/selection';
 import { ApplicationState } from './applicationReducer';
 import { createPage } from './pageReducer';
-import { ApplicationReducerContext, Primitives } from 'noya-state';
+import {
+  ApplicationReducerContext,
+  findIndexPath,
+  Primitives,
+} from 'noya-state';
 
 export type LayerAction =
   | [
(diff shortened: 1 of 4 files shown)


packages/noya-state/src/selectors/layerSelectors.ts
```diff
@@ -501,3 +501,15 @@ export function getConstrainedScaling(
       true)
   );
 }
+
+// Check if layer is visible in the page
+export function isLayerVisible(layer: Sketch.AnyLayer, canvasFrame: Rect) {
```
dabbott, line 506: I think we may want to use `getLayersInRect`, which will also handle transforms

https://github.com/noya-app/noya/blob/e3c896a128675123ec46608740f38d5f98767367/packages/noya-state/src/selectors/geometrySelectors.ts#L443-L452

packages/noya-state/src/reducers/layerReducer.ts
```diff
@@ -438,50 +439,89 @@ export function layerReducer(
...
+          ) {
+            // If the layer is on the current page and is visible
             addSiblingLayer(draftPage, indexPaths[0], newLayer);
-          } else if (!Layers.isPageLayer(selectedLayer)) {
-            addSiblingLayer(draftPage, selectedLayerIndexPath, newLayer);
+          } else if (!Layers.isPageLayer(parentLayer)) {
+            // If the selected layer is a parent layer but not a page layer
+            newLayer.frame = {
+              ...newLayer.frame,
+              x: parentLayer.frame.width / 2 - layerFrame.width / 2,
```
dabbott, line 487: To get a layer midpoint, we can use 

```ts
const parentBounds = getBounds(parentLayer.frame)
parentBounds.midX
parentBounds.midY
```

dabbott: And maybe a function that sets the midpoint of a frame would help?

Untested:

```ts
// noya-geometry rect.ts
export function moveRectCentered<T extends Rect>(rect: T, point: Point): T {
  const bounds = createBounds(rect);

  const delta: Point = {
    x: bounds.midX - point.x,
    y: bounds.midY - point.y,
  };

  return {
    ...rect,
    x: rect.x + delta.x,
    y: rect.y + delta.y,
  };
}
```

dabbott:
Looks good! Leave the comments, they're helpful 😁. Maybe just clean em up a little

It'll also be helpful to see how the tests are changed by these updates

dabbott (changes requested):

packages/noya-state/src/reducers/layerReducer.ts
milg15: The parents midx and y are in relationship to the canvas not internal x and y so it doesn't work

dabbott (approved):
The new behavior looks great! 

I merged with main and refactored the frame/bounds stuff a little, but thanks to the tests I think I didn't break anything 😆

Issue #290, opened 2021-11-30
Fix the size of the layer scaling & dragging overlay UI


dabbott:
While zoomed in:
- When scaling a layer, the UI that shows the dimensions, e.g. "30x30", is way too big
- When dragging a layer, the UI that shows "snap guides", e.g. the orange/purple lines and number indicating distance between layers, is way too big

I think both of these are in `SnapGuides.tsx`, and they mainly need to use `const zoom = useZoom();` to get the current zoom.

https://user-images.githubusercontent.com/1198882/144112671-c841a52a-55d6-4d94-9a35-196b7f57e452.mov

https://user-images.githubusercontent.com/1198882/144113190-bf5bf8a8-d430-4e9e-a2b7-2f76a44ed69b.mov

PR #292, opened 2021-12-04
keep size of snap guides #290

milg15:


dabbott:
With this approach the font renders kinda pixelated, since fonts typically render differently at very small sizes. I'm gonna push a change where we scale the whole text, instead of the individual properties

https://user-images.githubusercontent.com/1198882/144767299-601a41d4-5e8b-465d-b592-6f2647e0a9d8.mov

We also want the area displayed in the "A x B" to be the same, since it's measuring the area of the shape, and that isn't affect by zoom. That's small though, I'll update that too

dabbott:
You can try for these snapping guides and labels? I think we'll probably need a similar fix for the label, where we scale the whole thing. For the lines we can probably just adjust the thickness normally

https://user-images.githubusercontent.com/1198882/144767589-3f7c1784-0ba7-4fc2-bd9f-5e9ec3c36d88.mov

milg15:
I am not 100% sure how to position correctly the label :thinking: The affinity works weird

Issue #291, opened 2021-11-30
Support line caps and joins


dabbott:
Design tools like Sketch and Figma typically support adjustable "Line caps" and "Line joins". 

A "Line cap", or in our file format, `LineCapStyle`, is one of: `Butt`, `Round`, `Projecting` (aka Square)
A "Line join", or in our format, `LineJoinStyle`, is one of: `Miter`, `Round`, `Bevel`

These values are fairly standard across graphics libraries so I think they *should* be supported in CanvasKit.

### UI

We should update our "Borders" inspector UI to have an expandable *advanced* options section (expands via gear icon) that has 2 Select dropdowns: one for "Ends" (line caps) and one for "Angles" (line joins).

![Group 2](https://user-images.githubusercontent.com/1198882/144117104-a9108d57-486d-4947-856e-45c77cba02a6.png)

Note that the same line cap/join options apply to all borders if there are multiple.

### Implementation

- The UI state for "expanded" vs "collapsed" for the *advanced* section can be local to the borders inspector component, e.g. with useState. It does not need to persist when selecting a different layer
- We should show the existing data in the layer's border style properties (line cap & join) in the inspector
- We should have actions for updating the layer's border style properties
- We should render these line caps using CanvasKit, likely in `SketchShape.tsx`. This can come last, or could be a separate issue if it gets tricky.

### Example

Here's how this UI works in Sketch (which will be very similar to ours):

https://user-images.githubusercontent.com/1198882/144116257-c30fc59a-3474-47cc-8c3d-1ffcff63597a.mov

And Figma:

https://user-images.githubusercontent.com/1198882/144116589-57f5aad9-cc17-48b3-8c5a-6d86b9503160.mov

PR #293, opened 2021-12-04
Line caps and joins #291

milg15:


dabbott:
I pushed an update that gets the canvas rendering working. The main thing was propagating the border options into `getStrokedPath`, rather than setting them on the Paint. We actually convert every stroke into a new path and then "fill" it, since that lets us do things like gradient strokes.

packages/app/src/containers/BorderInspector.tsx
```diff
@@ -28,12 +42,81 @@ export default memo(function BorderInspector() {
...
+    [dispatch],
+  );
+  const onChangeLineJoinStyle = useCallback(
+    (value: string) => {
+      dispatch('setBorderLineJoin', parseInt(value));
+    },
+    [dispatch],
+  );
+
+  const BorderInspectorExpanded = useMemo(() => {
```
dabbott, line 72: Lowercase `borderInspectorExpanded` cause it's not a component

packages/app/src/components/inspector/ArrayController.tsx
```diff
@@ -30,10 +30,13 @@ interface ArrayControllerProps<Item> {
   title: ReactNode;
   sortable?: boolean;
   reversed?: boolean;
+  expanded?: boolean;
+  expandedContent?: ReactNode;
```
dabbott, line 34: For consistency with `renderItem` let's make this a "render prop" . I.e. `renderExpandedContent?: () => ReactNode`.

This also has the small perf benefit of not needing to run unless it's actually expanded and visible.

dabbott (changes requested):
Looks great! A commented on a couple small things. 

FYI I also remembered I removed the `borderOptions.isEnabled` check in my previous change. I don't think there's actually a way to disable border options in Sketch, and I don't think we'll have one either

milg15:
Hopefully, this makes sense

dabbott:
We were still using `expandedContent` on the CheckboxArrayController, so I swapped it for `renderExpandedContent`

dabbott (approved):

PR #301, opened 2022-01-24
Refactor noya-state to be mobile friendly

michalsek:


packages/noya-file-format/src/types.ts
```diff
@@ -293,6 +293,17 @@ export enum PatternFillType {
   Stretch = 2,
   Fit = 3,
 }
+/**
+ * Defines a pattern
+ */
+export type Pattern = {
```
dabbott, line 299: This file is auto-generated, so probably this type should go elsewhere. But I can move it later, not super important. Maybe we could do a separate `generated-types.ts` and `types.ts` can import it

dabbott (approved):

PR #310, opened 2022-02-22
Improve performance, fix color parser in rn canvaskit

michalsek:
- fixed SketchFileRenderer canvas transform and screen transform groups causing some of the components (like grid pixels) to be created on each render
- fixed color parser in react-native canvaskit class
- commented out mobile fps counter as it generated a lot of noise in profiling tools

### DIFF
--- a/packages/app-mobile/src/containers/CanvasRenderer.tsx
+++ b/packages/app-mobile/src/containers/CanvasRenderer.tsx
@@ -1,4 +1,5 @@
 import React, { memo } from 'react';
+import { StyleSheet } from 'react-native';
 import { ThemeProvider, useTheme } from 'styled-components';
 import { Canvas as SkiaCanvas } from '@shopify/react-native-skia';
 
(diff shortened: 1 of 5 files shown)


packages/app-mobile/src/containers/CanvasRenderer.tsx
```diff
@@ -20,7 +20,7 @@ const CanvasRenderer = () => {
   const theme = useTheme();
 
   return (
-    <StyledCanvas>
+    <SkiaCanvas style={{ flex: 1 }}>
```
dabbott, line 23: Since we're thinking about perf, move the style out of inline?

packages/app-mobile/src/containers/Toolbar.tsx
```diff
@@ -53,44 +57,54 @@ const Toolbar: React.FC<ToolbarProps> = (props) => {
...
-    if (!decodedImage) {
-      return;
-    }
+  const processImage = useCallback(
+    (base64Image: string, uri: string) => {
+      const data = base64ToArrayBuffer(base64Image);
+      const decodedImage = CanvasKit.MakeImageFromEncoded(data);
+      const extension =
+        uri.substring(uri.lastIndexOf('.')) === 'jpg' ? 'jpg' : 'png';
+      const name = uri.substring(uri.lastIndexOf('/'), uri.lastIndexOf('.'));
```
dabbott, line 86: There's a noya util `delimitedPath` that might simplify this string manipulation

dabbott (approved):
Looks great!

PR #321, opened 2022-03-24
[Workspace UI] Inspectors

michalsek:


packages/noya-utils/src/Platform.ts
```diff
@@ -0,0 +1,31 @@
...
+  Electron = 'Electron',
+}
+
+export const Platform = {
+  isNative: false,
+  isWeb: false,
+  isElectron: false,
+};
+
+export function setPlaform(platformName: PlatformName) {
```
dabbott, line 13: Is this something that could be inferred somehow instead of having to set it manually?

E.g. we could export a `currentPlatform` different value from `.ts` and `.native.ts`

dabbott (approved):

PR #326, opened 2022-05-03
Remove temp noya-web-designsystem and noya-icons

michalsek:


dabbott:
I think the code looks good, but I'm noticing the new icons don't seem as crisp:

| Old | New |
| --- | --- |
| Screen Shot 2022-05-04 at 11 36 18 AM | Screen Shot 2022-05-04 at 11 38 56 AM |

The original icons were designed for 15x15, so they look great at that size, and worse at any other size. I do think we should try to get them looking good before swapping the web code over.

PR #330, opened 2022-05-05
[Mobile] key commands

michalsek:
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-05-10 at 19 27 36


packages/app-mobile/ios/KeyCommandRegistry/KeyCommandRegistry.swift
```diff
@@ -0,0 +1,245 @@
...
+//  noyamobile
+//
+//  Created by Devin Abbott on 5/4/22.
+//
+import Foundation
+import React
+import UIKit
+
+let Modifiers = ["Shift", "Meta", "Ctrl", "Alt"]
+let SpecialKeys = ["Escape", "End", "Delete", "Home", "PageUp", "PageDown", "ArrowUp", "ArrowDown", "ArrowLeft", "ArrowRight"]
```
dabbott, line 12: These could be enums, and we could `switch` on the enum. Not super important.

There's no great way to automatically convert enum <==> string in swift, but there are some tricks that get you pretty close. E.g. from https://stackoverflow.com/a/52148845, https://stackoverflow.com/a/64365354

```swift
// This can go in a separate file, e.g. CaseIterableExtensions.swift, to allow easier type conversions
extension CaseIterable {
  static func from(string: String) -> Self? {
    return Self.allCases.first { string == "\($0)" }
  }
  func toString() -> String { "\(self)" }
}

enum SpecialKey: String, CaseIterable {
  case escape = "Escape"
  case end = "End"
  // ...
}

// Parse
if let specialKey = SpecialKey.from(string: key) {
  // Found it!
} else {
  // Not a special key
}
```

dabbott (approved):
Looks great!

PR #333, opened 2022-05-23
[Mobile] Colorpicker and inspector fixes

michalsek:
- prevents opening multiple color pickers / any objects depending on Popover component
- Fixes issue with content of attribute inspector being duplicated when interacting with it


dabbott (approved):
We should probably also be able to close a popover by tapping outside of it. (doesn't have to be this PR)

With things like gradients, tapping specifically the canvas shouldn't close the popover/picker, but for most other things tapping anywhere else should close it

